PR #12, opened 2021-12-13
RPC API fixes

ray-kast:
Currently this is still a draft as bid timestamps are collected by spamming getProgramAccounts calls (which is not desirable).


crates/rpc/src/rpc.rs
```diff
@@ -95,199 +55,66 @@ impl Rpc for Server {
...
-                        name,
-                        uri,
-                    });
-                },
-            }
-        }
-
-        Ok(listings.into_values().collect())
+        queries::listings_triple_join(|q| {
+            q.select(queries::LISTINGS_TRIPLE_JOIN_COLS)
```
kristianeboe, line 59: How do I access ::listings from this context? to apply a filter based on a listing f.ex

PR #13, opened 2021-12-14
feat: get featured listings

kristianeboe:
Hey can we add this? 

Simple offset and limit for featured and non featured listings + suggestions for additional filters

### DIFF
--- a/crates/rpc/src/rpc.rs
+++ b/crates/rpc/src/rpc.rs
@@ -23,6 +23,8 @@ fn internal_error<E: Into<indexer_core::error::Error>>(
 
 #[rpc]
 pub trait Rpc {
+    #[rpc(name = "getFeaturedListings")]
+    fn get_featured_listings(&self) -> Result<Vec<Listing>>;
     #[rpc(name = "getListings")]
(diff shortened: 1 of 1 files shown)


crates/rpc/src/rpc.rs
```diff
@@ -52,10 +54,17 @@ impl Server {
...
+            .map_err(internal_error("Failed to load listings"))
+    }
+
     fn get_listings(&self) -> Result<Vec<Listing>> {
         let db = self.db()?;
 
-        listings_triple_join::load(|q| q, &db, Local::now().naive_utc())
+        listings_triple_join::load(|q| q.offset(4), &db, Local::now().naive_utc())
             .map_err(internal_error("Failed to load listings"))
     }
```
ray-kast, line 69: I think it might be good to rename this method to something like `getNonFeaturedListings` for clarity.  Since the original `getListings` was kind of a stopgap, I'd be fine with removing it altogether if we're no longer using it.

ray-kast (changes requested):

ray-kast:
Closing for now, I'll probably resurrect this as part of the pagination ticket

PR #18, opened 2021-12-21
Add a storefront denylist.

ray-kast:
Current implementation tracks owner pubkeys and remove any associated storefronts from RPC queries.

We should be certain that this is watertight before merging.

### DIFF
--- a/crates/core/src/db/queries/listings_triple_join.rs
+++ b/crates/core/src/db/queries/listings_triple_join.rs
@@ -16,6 +16,7 @@ use diesel::{
 };
 use solana_sdk::native_token::sol_to_lamports;
 
+use super::store_denylist::{owner_address_ok, OwnerAddressOk};
 use crate::{
     db::{
         models::ListingsTripleJoinRow,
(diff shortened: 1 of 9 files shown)


crates/core/src/db/queries/listings_triple_join.rs
```diff
@@ -58,10 +63,13 @@ pub fn rejected(now: NaiveDateTime) -> Rejected {
...
-    listings::ended.eq(false).and(not(already_ended
-        .or(ended_instant_sale)
-        .or(not_ending_soon)
-        .or(too_expensive)))
+    listings::ended
+        .eq(false)
+        .and(not(already_ended
+            .or(ended_instant_sale)
+            .or(not_ending_soon)
+            .or(too_expensive)))
```
kespinola, line 71: love how this reads.

kespinola (approved):

crates/core/src/db/queries/listings_triple_join.rs
ray-kast: Yeah, breaking that up into variables ended up making the query scan cleaner in addition to making writing the resulting DSL type easier.

kristianeboe (approved):
Amazing work!

PR #19, opened 2021-12-22
FULL TEXT SEARCH

ray-kast:
Due to the ordering of the diesel migrations, this shouldn't be merged until #18 is ready, but that shouldn't be a big deal given the ticket priorities.

### DIFF
--- a/crates/rpc/src/rpc.rs
+++ b/crates/rpc/src/rpc.rs
@@ -2,14 +2,14 @@ use indexer_core::db::{
     models,
     queries::{listings_triple_join, metadata_edition, store_denylist},
     tables::{bids, listing_metadatas, listings, metadata_creators, metadatas, storefronts},
-    Pool, PooledConnection,
+    websearch_to_tsquery, Pool, PooledConnection,
 };
 use jsonrpc_core::{Error, Result};
 use jsonrpc_derive::rpc;
 
 use crate::{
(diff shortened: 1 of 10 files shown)


crates/rpc/src/rpc.rs
```diff
@@ -27,6 +27,8 @@ pub trait Rpc {
     fn get_listings(&self) -> Result<Vec<Listing>>;
     #[rpc(name = "getStorefronts")]
     fn get_storefronts(&self) -> Result<Vec<Storefront>>;
+    #[rpc(name = "searchStorefronts")]
+    fn search_storefronts(&self, query: String) -> Result<Vec<Storefront>>;
```
kespinola, line 31: Thoughts on moving search to a filter option on get storefronts?

`getStorefronts(search:)`

If search option is set then add the filter doing tsearch.

kespinola (changes requested):

crates/rpc/src/rpc.rs
ray-kast: I can look into that, my only concern was whether that would break the API by changing parameter counts.  If it requires a null to be explicitly added to the params list then we can't touch that until we refactor endpoints for v2, but if not I can do it right away.

kespinola (approved):

PR #22, opened 2021-12-27
Run migrations on every deploy

kespinola:
### Changes
- Add diesel cli to image
- Include release step in heroku file for running migrations


ray-kast (changes requested):
The binaries run migrations when they start, diesel-cli is not required.

PR #24, opened 2022-01-03
Index bidder metadata for auctions.

ray-kast:
This PR includes a separate job graph entry point for the indexer backend that collates bid metadata from a single getProgramAccounts call in parallel to the auction cache search.  This data can then be drawn from to store bid info when indexing auctions.


ray-kast:
I think it's ready for final review! @dandlezzz

dandlezzz (approved):
wow

PR #28, opened 2022-01-06
add arweave block timestamp as updated_at for storefronts

dandlezzz:
As precise as arweave allows.
addresses #27

### DIFF
--- a/crates/indexer/src/get_storefronts.rs
+++ b/crates/indexer/src/get_storefronts.rs
@@ -48,12 +48,18 @@ struct QueryNode {
     #[allow(dead_code)]
     owner: QueryOwner,
     tags: Vec<QueryTag>,
+    block: Timestamp,
 }
 
 #[derive(Debug, Deserialize)]
 struct QueryOwner {
     #[allow(dead_code)]
     address: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct Timestamp {
+    timestamp: i64,
 }
 
 #[derive(Debug, Deserialize)]
@@ -84,14 +90,15 @@ const QUERY: &str = r#"query GetStorefronts($after: String, $first: Int) {
 
             node {
                 id
-
                 owner {
                     address
                 }
-
                 tags {
                     name
                     value
+                }
+                block {
+                    timestamp
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


crates/indexer/src/get_storefronts.rs
```diff
@@ -129,6 +138,8 @@ fn process_tags(
         let logo_url = tags
             .remove("holaplex:theme:logo:url")
             .unwrap_or_else(String::new);
+        
+        let created_at = Some(timestamp).map(|t| NaiveDateTime::from_timestamp(t, 0));
```
ray-kast, line 142:
```suggestion
        let created_at = Some(NaiveDateTime::from_timestamp(timestamp, 0));
```

ray-kast: `Some(...).map(...)` is redundant, you can just operate on `timestamp` directly and wrap the whole thing in an `Option` afterwards

crates/rpc/src/rpc.rs
```diff
@@ -76,13 +76,15 @@ impl Rpc for Server {
...
                      logo_url,
+                     created_at,
                  }| RpcStorefront {
                     owner_address: owner_address.into_owned(),
                     subdomain: subdomain.into_owned(),
                     title: title.into_owned(),
                     description: description.into_owned(),
                     favicon_url: favicon_url.into_owned(),
                     logo_url: logo_url.into_owned(),
+                    created_at,
```
ray-kast, line 87:
```suggestion
                    created_at: created_at.map(Timestamp::from_utc),
```

In accordance with the suggestion to `rpc_models.rs`

crates/rpc/src/rpc_models.rs
```diff
@@ -118,4 +118,5 @@ pub struct Storefront {
     pub description: String,
     pub favicon_url: String,
     pub logo_url: String,
+    pub created_at: Option<NaiveDateTime>,
```
ray-kast, line 121:
```suggestion
    pub created_at: Option<Timestamp>,
```

The `rpc_models` module uses a `Timestamp` struct to enforce consistent formatting of `NaiveDateTime` objects.

ray-kast (changes requested):
Looks good overall, just needs some minor tweaks and lints.

crates/indexer/src/get_storefronts.rs
```diff
@@ -56,6 +57,12 @@ struct QueryOwner {
     address: String,
 }
 
+#[derive(Debug, Deserialize)]
+struct Timestamp {
+    #[allow(dead_code)]
```
ray-kast, line 62:
```suggestion
```

Oh, forgot to mention - don't need to include this since this isn't dead code.

ray-kast (approved):

PR #43, opened 2022-01-09
Index NFTs by Creator, dead code for now :-)

### LINKED ISSUE #33: Index NFTs created by a wallet
Indexing sync based on creator address of NFTs. This results in a gPA matching on the first slot of creators list of the Metadata for the given address.

dandlezzz:
Index NFTs by creator. We need a web interface here, I'll go with the v1 RPC for now. A CLI might be helpful one day too.

resolves #33 

- [x] function to fetch nfts by first creator
- [x] Add creators wallet table, model
- [x] Threaded writes to database 

This is dead code for now, #45 handles the calling of these functions.

### DIFF
--- a/crates/core/migrations/2022-01-10-200616_create_requested_creators/down.sql
+++ b/crates/core/migrations/2022-01-10-200616_create_requested_creators/down.sql
@@ -1,2 +1,1 @@
--- This file should undo anything in `up.sql`
 drop table creators;
--- a/crates/core/migrations/2022-01-10-200616_create_requested_creators/down.sql
+++ b/crates/core/migrations/2022-01-10-200616_create_requested_creators/down.sql
@@ -1,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+drop table creators;
--- a/crates/core/migrations/2022-01-10-200616_create_requested_creators/up.sql
+++ b/crates/core/migrations/2022-01-10-200616_create_requested_creators/up.sql
@@ -1,5 +1,5 @@
 create table creators (
-  address bytea primary key not null,
-  created_at              timestamp not null default now(),
-  updated_at              timestamp null
+  address     bytea     primary key not null,
+  created_at  timestamp not null default now(),
+  updated_at  timestamp null
 );
--- a/crates/core/migrations/2022-01-10-200616_create_requested_creators/up.sql
+++ b/crates/core/migrations/2022-01-10-200616_create_requested_creators/up.sql
@@ -1,0 +1,5 @@
+create table creators (
+  address bytea primary key not null,
+  created_at              timestamp not null default now(),
+  updated_at              timestamp null
+);
--- a/crates/indexer/src/bits/metadata.rs
+++ b/crates/indexer/src/bits/metadata.rs
@@ -14,19 +14,19 @@ pub const MAX_NAME_LENGTH: usize = 32;
 pub const MAX_URI_LENGTH: usize = 200;
 pub const MAX_SYMBOL_LENGTH: usize = 10;
 pub const MAX_CREATOR_LEN: usize = 32 + 1 + 1;
-pub const FIRST_CREATOR_LENGTH: usize = 1 + // key
-32 + // update auth
-32 + // mint
-4 + // name string length
-MAX_NAME_LENGTH + // name
-4 + // uri string length
-MAX_URI_LENGTH + // uri*
-4 + // symbol string length
-MAX_SYMBOL_LENGTH + // symbol
-2 + // seller fee basis points
-1 + // whether or not there is a creators vec
-4 +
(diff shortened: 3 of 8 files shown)


crates/indexer/src/bits/metadata.rs
```diff
@@ -1,11 +1,50 @@
...
+4 + // name string length
+MAX_NAME_LENGTH + // name
+4 + // uri string length
+MAX_URI_LENGTH + // uri*
+4 + // symbol string length
+MAX_SYMBOL_LENGTH + // symbol
+2 + // seller fee basis points
+1 + // whether or not there is a creators vec
+4 +
+0 * MAX_CREATOR_LEN; // creators vec length
```
ray-kast, line 29: Can any of these constants be imported from the metaplex crates? We already pull them in as dependencies.

dandlezzz: @ray-kast not seeing these values as exports, any ideas where they might be?

ray-kast: https://docs.rs/metaplex-token-metadata/0.0.1/metaplex_token_metadata/state/index.html#constants looks like they’re exported from `metaplex_token_metadata::state`

crates/core/migrations/2022-01-10-200616_create_requested_creators/up.sql
```diff
@@ -0,0 +1,5 @@
+create table creators (
+  address bytea primary key not null,
+  created_at              timestamp not null default now(),
+  updated_at              timestamp null
```
ray-kast, line 4:
```suggestion
  address     bytea     primary key not null,
  created_at  timestamp not null default now(),
  updated_at  timestamp null
```

crates/core/migrations/2022-01-10-200616_create_requested_creators/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
```
ray-kast, line 1:
```suggestion
```

crates/indexer/src/bits/metadata.rs
```diff
@@ -1,11 +1,50 @@
...
+    },
+    pubkeys,
 };
-use metaplex_token_metadata::state::Metadata as MetadataAccount;
+use metaplex_token_metadata::state::{
+    Metadata as MetadataAccount, MAX_NAME_LENGTH, MAX_SYMBOL_LENGTH, MAX_URI_LENGTH,
+};
+
+use crate::{client::prelude::*, prelude::*, util, Client, EditionKeys, Job, ThreadPoolHandle};
+pub const FIRST_CREATOR_LENGTH: usize =
```
ray-kast, line 14:
```suggestion
pub const FIRST_CREATOR_OFFSET: usize =
```

crates/indexer/src/bits/metadata.rs
```diff
@@ -1,11 +1,50 @@
...
+fn get_metadatas_by_primary_creator(
+    client: &Client,
+    creator_address: Pubkey,
+) -> Result<
+    Vec<(solana_sdk::pubkey::Pubkey, solana_sdk::account::Account)>,
+    solana_client::client_error::ClientError,
+> {
+    client.get_program_accounts(pubkeys::metadata(), RpcProgramAccountsConfig {
+        filters: Some(vec![RpcFilterType::Memcmp(Memcmp {
+            offset: FIRST_CREATOR_LENGTH,
```
ray-kast, line 26:
```suggestion
            offset: FIRST_CREATOR_OFFSET,
```

crates/indexer/src/entry.rs
```diff
@@ -144,6 +146,9 @@ fn create_pool(
                 Job::GetStorefronts => get_storefronts::run(&client, handle),
                 Job::GetBidderMetadata => bidder_metadata::get(&client, &bid_map, handle),
                 Job::GetBidderMetadataSolo => bidder_metadata::get_solo(&client, handle),
+                Job::MetadataByCreator(pubkey) => {
+                    metadata::handle_metadata_by_creator(&client, pubkey, handle)
+                },
```
ray-kast, line 151: This should go below the match arm for `Job::SoloBidsForAuction` for consistency with the enum definition order

ray-kast (changes requested):
:pleading_face:

crates/core/migrations/2022-01-10-200616_create_requested_creators/up.sql
ray-kast: Actually, additional note on this - do we need the migration if we're saving this table for later?

ray-kast (approved):

dandlezzz:
we are going to do this through rabbitmq

PR #44, opened 2022-01-10
Add Graphql Server

kespinola:
### Changes
- Add new graphql service to act as the HTTP public interface into Marketplace resources such as NFTs, marketplaces, and orders.
- Update readme to account for starting graphql server.

resolves https://github.com/holaplex/metaplex-indexer/issues/37


README.md
```diff
@@ -32,19 +38,24 @@ To run the indexer, simply enter the repository root and run:
 $ cargo run --bin metaplex-indexer
 ```
 
-## Running `rpc`
+## Running HTTP Servers
+
+If port `3000` is already in use the `PORT` environment variable can be used chang the listener port of the servers:
```
ray-kast, line 43: The server actually uses a command line argument that defaults to pulling from the `PORT` env var, we should probably mention this.  Additionally, all binaries should have a `--help` option we can mention that explains all this as well.

README.md
```diff
@@ -32,19 +38,24 @@ To run the indexer, simply enter the repository root and run:
...
-### If port `3000` is already in use
+### `graph`
 
-The `PORT` environment variable can be set to change the port `rpc` listens on:
+To run the GraphQL Server, execute the following command:
 
 ```sh
-$ PORT=3001 cargo run --bin metaplex-indexer-rpc
+$ cargo run --bin metaplex-graph-server
 ```
```
ray-kast, line 59: I would prefer if this were in `crates/graphql` and named `metaplex-indexer-graphql` for consistency

crates/graph/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+                    .max_age(3600)
+                    .finish(),
+            )
+            .service(web::resource("/").route(web::post().to(graphql)))
+            .service(web::resource("/graphiql").route(web::get().to(graphiql)))
+    })
+    .bind("127.0.0.1:3000")?
+    .run()
+    .await
+}
```
ray-kast, line 61: This should pull in `clap` and have a similar argument parser to the `rpc` crate for allowing the port to be changed.  Additionally I have already set up of some of the boilerplate written here in the `indexer_core` crate.  I can look into making sure it works with async, but I would much prefer to reuse that than pulling in the same dependencies in a slightly different manner.

README.md
```diff
@@ -17,10 +24,9 @@ Then:
 $ cargo install diesel_cli --no-default-features --features postgres
 ```
 
-Once you have the requisite dependencies, you can get set up by running:
+Once you have the required dependencies, you can get get developing by running the datastore for the indexer within a container in the background. After the Postgres DB is running start whichever serves you are looking to develop on:
```
ray-kast, line 27: `./start-developing.sh` sets up docker-compose and diesel.  Because of how diesel works this actually needs to be run before one can start developing, not just before they can start a server.

crates/graph/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "metaplex-graph-server"
+version = "0.1.0"
+authors = ["Holaplex Engineering <hola@holaplex.com>"]
+edition = "2021"
```
ray-kast, line 5: The metadata will need to be updated in order for this crate to be publishable.

crates/graph/src/main.rs
```diff
@@ -0,0 +1,61 @@
+//! Actix web juniper example
+//!
+//! A simple example integrating juniper in actix-web
```
ray-kast, line 3:
```suggestion
//! GraphQL server to read from `metaplex-indexer`

#![deny(
    clippy::suspicious,
    clippy::style,
    missing_debug_implementations,
    missing_copy_implementations
)]
#![warn(clippy::pedantic, clippy::cargo, missing_docs)]
```

ray-kast (changes requested):
The `graphql` crate needs some consistency fixes to line up with the other crates.

ray-kast (approved):

PR #51, opened 2022-01-14
RABBITMQ

ray-kast:


crates/accountsdb-rabbitmq/src/selectors.rs
```diff
@@ -0,0 +1,54 @@
...
+            .map(|s| bs58::decode(s).into_vec().map(Vec::into_boxed_slice))
+            .collect::<Result<_, _>>()
+            .context("Failed to parse account owner keys")?;
+
+        Ok(Self { owners })
+    }
+
+    #[inline]
+    pub fn is_selected(&self, acct: &ReplicaAccountInfo) -> bool {
+        self.owners.is_empty() || self.owners.contains(acct.owner)
```
dandlezzz, line 25: I don't think this default is the right approach for us. Can we switch to filtering all of them out if the user specifies none. Bringing explicit support for all via config is a good future goal.

PR #54, opened 2022-01-18
Index token accounts

dandlezzz:
addresses #46 

By indexing token accounts we know who holds which mints and how many. We are currently limiting this work to token accounts where we know the mint is a currently indexed NFT. We will likely have to return to this later. There could be a race condition where an NFT is created and we miss its token account. Will address in future PRs.

### DIFF
--- a/crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
+++ b/crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
@@ -1,7 +1,7 @@
 create table token_accounts (
-  address varchar(48) primary key not null,
-  mint_address varchar(48) not null,
+  address       varchar(48) primary key not null,
+  mint_address  varchar(48) not null,
   owner_address varchar(48) not null,
-  amount bigint default 0,
-  updated_at timestamp not null
-)
+  amount        bigint      default 0,
+  updated_at    timestamp   not null
+);
--- a/crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
+++ b/crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
@@ -3,6 +3,5 @@ create table token_accounts (
   mint_address varchar(48) not null,
   owner_address varchar(48) not null,
   amount bigint default 0,
-  created_at timestamp not null,
-  updated_at timestamp
+  updated_at timestamp not null
 )
--- a/crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
+++ b/crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
@@ -1,0 +1,8 @@
+create table token_accounts (
+  address varchar(48) primary key not null,
+  mint_address varchar(48) not null,
+  owner_address varchar(48) not null,
+  amount bigint default 0,
+  created_at timestamp not null,
+  updated_at timestamp
+)
--- a/crates/core/src/db/models.rs
+++ b/crates/core/src/db/models.rs
@@ -9,7 +9,7 @@ use chrono::NaiveDateTime;
 
 use super::schema::{
     bids, editions, listing_metadatas, listings, master_editions, metadata_creators, metadatas,
-    storefronts, token_accounts
+    storefronts, token_accounts,
 };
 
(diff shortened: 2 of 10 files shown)


crates/core/src/db/models.rs
```diff
@@ -124,6 +124,22 @@ pub struct MetadataCreator<'a> {
...
+#[derive(Debug, Clone, Queryable, Insertable, AsChangeset)]
+pub struct TokenAccount<'a> {
+    /// The address of this account
+    pub address: Cow<'a, str>,
+    /// The mint address of the token
+    pub mint_address: Cow<'a, str>,
+    /// The owner token
+    pub owner_address: Cow<'a, str>,
+    /// The amount of the token, often 1
+    pub amount: i64,
```
ray-kast, line 138: is `amount` immutable?

crates/core/migrations/2022-01-18-164927_create_token_accounts_table/up.sql
```diff
@@ -0,0 +1,7 @@
+create table token_accounts (
+  address varchar(48) primary key not null,
+  mint_address varchar(48) not null,
+  owner_address varchar(48) not null,
+  amount bigint default 0,
+  updated_at timestamp not null
+)
```
ray-kast, line 7:
```suggestion
create table token_accounts (
  address       varchar(48) primary key not null,
  mint_address  varchar(48) not null,
  owner_address varchar(48) not null,
  amount        bigint      default 0,
  updated_at    timestamp   not null
);
```

crates/indexer/src/bits/token_account.rs
```diff
@@ -0,0 +1,50 @@
...
+use spl_token::state::Account as TokenAccount;
+
+use crate::{prelude::*, Client};
+
+pub fn process(client: &Client, pubkey: Pubkey, token_account: TokenAccount) -> Result<()> {
+    let mint = token_account.mint.to_string();
+    let owner = token_account.owner.to_string();
+    let pubkey = pubkey.to_string();
+    let db = client.db()?;
+    let now = Local::now().naive_utc();
```
ray-kast, line 20:
```suggestion
    let mint = token_account.mint.to_string();
    let db = client.db()?;
```

Move some of the string conversions after the short-circuit arm.

crates/indexer/src/bits/token_account.rs
```diff
@@ -0,0 +1,50 @@
...
+    if !select(exists(
+        metadatas::table.filter(metadatas::mint_address.eq(&mint)),
+    ))
+    .get_result(&db)
+    .context("Failed to check mint address for existing mint")?
+    {
+        return Ok(());
+    }
+
+    let amount = token_account.amount as i64;
```
ray-kast, line 31:
```suggestion
    let pubkey = pubkey.to_string();
    let amount = token_account.amount as i64;
    let owner = token_account.owner.to_string();
    let now = Local::now().naive_utc();
```

ray-kast (changes requested):
no cardinal sins here!

crates/core/src/db/models.rs
dandlezzz: I believe it is mutable.

ray-kast: will we need to reindex if amount changes, or is that not something we care about? @dandlezzz

crates/indexer/src/bits/token_account.rs
```diff
@@ -0,0 +1,50 @@
...
+        metadatas::table.filter(metadatas::mint_address.eq(&mint)),
+    ))
+    .get_result(&db)
+    .context("Failed to check mint address for existing mint")?
+    {
+        return Ok(());
+    }
+
+    let pubkey = pubkey.to_string();
+    let amount = token_account.amount as i64;
```
ray-kast, line 29:
```suggestion
    let amount: i64 = token_account.amount.try_into().context("Token amount was too big to store")?;
```

ray-kast (changes requested):
Somehow glossed over a lint warning on my first review, whoops.  The change I suggested will probably trigger a format CI error, you may want to just make the change locally and `cargo fmt`.

ray-kast (approved):

PR #61, opened 2022-01-20
GraphQL - Backend to combine creators with NFT metadata before return to client

echohtp:


ray-kast (changes requested):
Hey, I changed the CI flow, to get CI builds working again you'll need to merge or rebase onto dev

kespinola (changes requested):
This is where you want to use a resolver (and then data loader). The nft query should be what its is now but the object be extended with a creators resolver. This is where you make the query for the creators of the nfts.

https://github.com/rkudryashov/graphql-rust-demo/blob/master/planets-service/src/graphql.rs

I'm going to put something together according to the above. Lets sync on this.

PR #72, opened 2022-01-24
Ryans/optimize dockerfile

dandlezzz:
wip, working on deployments


ray-kast (changes requested):
Can we merge the revert first and not include it in this PR?

ray-kast (approved):

PR #78, opened 2022-01-24
Revert "Storefronts Graphql"

kespinola:
Reverts holaplex/metaplex-indexer#69


dandlezzz (approved):
will be back one day 😄

PR #79, opened 2022-01-24
NFT Details

kespinola:
## Changes
Metadata from URI is parsed and inserted into database
tables created:
metadata_jsons
attributes
files
metadata_collections

attributes, files and metadata_collections tables may have NULL values for attributes other than metadata_address and Diesel doesn't allow tables without primary key so had to insert an id variable as primary key.

- Query the contents of the nft metadata json through the nft details field.
- Use use a data loader to query metadata json in 1 query given a desired list of nfts.
- Fix up model layout for metadata json.
- Move graphql server to execute with async.

Screen Shot 2022-01-24 at 7 45 25 PM


crates/graphql/Cargo.toml
```diff
@@ -16,7 +16,10 @@ categories = ["cryptocurrency::cryptocurrencies", "web-programming"]
 [dependencies]
 actix-cors = "0.4.0"
 actix-web = "3.3.3"
+dataloader = "0.14.0"
 juniper = "0.15"
+futures = "0.3"
+async-trait = "0.1"
```
ray-kast, line 22:
```suggestion
actix-cors = "0.4.0"
actix-web = "3.3.3"
async-trait = "0.1"
dataloader = "0.14.0"
juniper = "0.15"
```

ray-kast (changes requested):
`futures` is likely an extraneous import, generally `std::future` has the necessary items

crates/graphql/src/main.rs
```diff
@@ -39,19 +41,22 @@ async fn graphql(
...
 ) -> Result<HttpResponse, Error> {
-    let json = web::block(move || {
-        let res = data.execute_sync(&st, &());
-        serde_json::to_string(&res)
-    })
-    .await?;
+    let db_pool =
+    db::connect(db::ConnectMode::Read).context("Failed to connect to Postgres").unwrap();
+
+    let db_connection = db_pool.get().unwrap();
```
kespinola, line 47: use closure to pass in db_pool. reference db_connection from context in queries instead of pulling form instance of query root.

crates/graphql/src/schema.rs
```diff
@@ -88,17 +121,74 @@ impl<'a> From<models::Storefront<'a>> for Storefront {
...
+        {
+            let desc = match description {
+                Some(val) => val.into_owned(),
+                None => "".to_string(),
+            };
+
+            let img = match image {
+                Some(val) => val.into_owned(),
+                None => "".to_string(),
+            };
```
kespinola, line 161: Anyway to do this cleaner?

crates/graphql/src/schema.rs
```diff
@@ -88,17 +121,74 @@ impl<'a> From<models::Storefront<'a>> for Storefront {
...
+
+#[async_trait]
+impl BatchFn<String, Option<NftDetail>> for NftDetailBatcher {
+    async fn load(&mut self, keys: &[String]) -> HashMap<String, Option<NftDetail>> {
+        let conn = &self.db_connection;
+        let mut hash_map = HashMap::new();
+
+        for key in keys {
+            hash_map.insert(key.clone(), None);
+        }
```
kespinola, line 140: While building this out for John Le there was 1 NFT that didn't have an associated NFT so the load call would panic. This is in t default every key to None and then override. Now that I review thinking the best is to capture the panic in the load and return none there. Going to look into that any start on capturing the error welcome.

crates/core/src/db/models.rs
```diff
@@ -226,3 +211,81 @@ pub struct ListingsTripleJoinRow {
...
+    pub animation_url: Option<String>,
+    /// Metadata External Url
+    pub external_url: Option<String>,
+    /// Metadata Category
+    pub category: Option<String>,
+    /// Metadata Timestamp
+    pub updated_at: Option<NaiveDateTime>,
+    /// Metadata URI raw json
+    pub raw_content: Option<Cow<'a, serde_json::Value>>,
+}
```
kespinola, line 237: If you get a large type mismatch from trying to select columns check to make sure that this model has the columns defined in the same order as the schema.rs and the types are the same. In this case the null values needed tom additional options and reorder. @imabdulbasit

crates/graphql/Cargo.toml
kespinola: yep. still works without 👍

crates/graphql/src/schema.rs
echohtp: I _**think**_ you can do this

```
let desc = description.into_owned().ok_or("");
let img = image.into_owned().ok_or("");
...
```

crates/core/migrations/2022-01-22-003337_create_metadata_json/up.sql
```diff
@@ -0,0 +1,41 @@
...
+
+create table metadata_collections
+(
+    metadata_address varchar(48) not null,
+    name text,
+    family text,
+    id SERIAL PRIMARY KEY
+);
+
+create index metadata_jsons_metadata_address_index ON metadata_jsons (metadata_address);
```
ray-kast, line 41:
```suggestion
create table metadata_jsons (
  metadata_address  varchar(48) primary key not null,
  fingerprint       bytea,
  description       text,
  image             text,
  animation_url     text,
  external_url      text,
  category          text,
  updated_at        timestamp,
  raw_content       jsonb
);

create table attributes (
  metadata_address  varchar(48) not null,
  name              text,
  value             text,
  trait_type        text,
  id                uuid primary key default gen_random_uuid()
);

create table files (
  metadata_address  varchar(48) not null,
  uri               text,
  file_type         text,
  id                uuid primary key default gen_random_uuid()
);

create table metadata_collections (
  metadata_address  varchar(48) not null,
  name              text,
  family            text,
  id                uuid primary key default gen_random_uuid()
);

create index metadata_jsons_metadata_address_index
on metadata_jsons (metadata_address);
```

crates/core/src/db/models.rs
```diff
@@ -224,3 +209,81 @@ pub struct ListingsTripleJoinRow {
...
+    pub address: Cow<'a, str>,
+    /// The mint address of the token
+    pub mint_address: Cow<'a, str>,
+    /// The owner token
+    pub owner_address: Cow<'a, str>,
+    /// The amount of the token, often 1
+    pub amount: i64,
+    /// updated_at
+    pub updated_at: NaiveDateTime,
+}
```
ray-kast, line 289: Why does this get moved here?  Makes the diff confusing

crates/graphql/src/schema.rs
```diff
@@ -1,15 +1,24 @@
...
+
+#[derive(Debug, Clone, GraphQLObject)]
+struct NftDetail {
+    description: String,
+    image: String,
+}
 
-#[derive(GraphQLObject)]
-#[graphql(description = "A Solana NFT")]
+#[derive(Debug, Clone)]
```
ray-kast, line 21:
```suggestion
#[derive(Debug, Clone)]
#[allow(dead_code)]
```

to silence the compiler warnings

crates/indexer/src/bits/mod.rs
```diff
@@ -4,5 +4,5 @@ pub mod bidder_metadata;
 pub mod edition;
 pub mod get_storefronts;
 pub mod metadata;
+pub mod metadata_uri;
 pub mod store_owner;
-pub mod token_account;
```
ray-kast, line 8: Why is this removed in this PR?  Is this a diff error, or do we need to move some embedded `TokenAccount` refactor to its own PR?

imabdulbasit: It was added in my PR

crates/graphql/src/main.rs
```diff
@@ -35,38 +35,47 @@ fn graphiql(uri: String) -> impl Fn() -> HttpResponse + Clone {
...
+    db_pool: &Arc<Pool>,
+) -> impl Fn(
+    web::Data<Arc<Schema>>,
+    web::Json<GraphQLRequest>,
+) -> Pin<Box<dyn Future<Output = Result<HttpResponse, Error>> + Send>>
++ Clone {
+    let db_pool = Arc::clone(db_pool);
+
+    move |st, data| {
+        let db_connection = db_pool.get().unwrap();
```
kespinola, line 48: @ray-kast do we not need to close the connection after getting the res? or is that handled with the function exiting?

crates/graphql/src/main.rs
```diff
@@ -35,38 +35,47 @@ fn graphiql(uri: String) -> impl Fn() -> HttpResponse + Clone {
...
+            Arc::new(db::connect(db::ConnectMode::Read).context("Failed to connect to Postgres")?);
 
         let mut addr: SocketAddr = "0.0.0.0:3000".parse().unwrap();
         addr.set_port(port);
         info!("Listening on {}", addr);
 
         let graphiql_uri = format!("http://{}", addr);
 
-        let schema = std::sync::Arc::new(schema::create(db_conn));
+        let schema = Arc::new(schema::create(db_pool.clone()));
```
kespinola, line 78: Note to refactor attaching db_pool to the query root. Instead start reading the connection produced for the request.

ray-kast (dismissed):

ray-kast (approved):

PR #81, opened 2022-01-25
Revert "Revert "Storefronts Graphql""

kespinola:
Reverts holaplex/metaplex-indexer#78

### DIFF
--- a/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/down.sql
+++ b/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/down.sql
@@ -1,0 +1,3 @@
+-- This file should undo anything in `up.sql`
+alter table storefronts
+drop column banner_url;
--- a/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/down.sql
+++ b/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/down.sql
@@ -1,3 +1,2 @@
--- This file should undo anything in `up.sql`
 alter table storefronts
 drop column banner_url;
--- a/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
+++ b/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
@@ -1,0 +1,3 @@
+-- Your SQL goes here
+alter table storefronts
+add column banner_url text not null;
--- a/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
+++ b/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
@@ -1,3 +1,3 @@
 -- Your SQL goes here
 alter table storefronts
-add column banner_url text not null;
+add column banner_url text not null default '';
--- a/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
+++ b/crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
@@ -1,3 +1,2 @@
--- Your SQL goes here
 alter table storefronts
-add column banner_url text not null default '';
+add column banner_url text null;
--- a/crates/graphql/src/schema.rs
+++ b/crates/graphql/src/schema.rs
@@ -1,12 +1,12 @@
 use indexer_core::{
     db::{
         models,
-        tables::{metadata_creators, metadatas},
+        tables::{metadata_creators, metadatas, storefronts},
         Pool,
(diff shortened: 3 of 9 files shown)


crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+alter table storefronts
+add column banner_url text not null default '';
```
kespinola, line 3: change to default column to '' string.

crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/down.sql
```diff
@@ -0,0 +1,3 @@
+-- This file should undo anything in `up.sql`
```
ray-kast, line 1:
```suggestion
```

crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
```
ray-kast, line 1:
```suggestion
```

crates/core/migrations/2022-01-21-232334_add_column_storefronts_banner_url/up.sql
ray-kast: Is this preferable to dropping the `NOT NULL` constraint?  I think having the column be nullable (at least as a temporary measure) would cause fewer logic errors.

ray-kast (changes requested):
RE: changing the Rust version - the entire project plus CI is configured to use a very specific nightly build of 1.58.  If we need to use another version for the AccountsDB plugin we should configure the build process for only that crate.

crates/graphql/src/schema.rs
```diff
@@ -51,6 +51,43 @@ impl<'a> From<models::Metadata<'a>> for Nft {
...
+        }: models::Storefront,
+    ) -> Self {
+        Self {
+            owner_address: owner_address.into_owned(),
+            subdomain: subdomain.into_owned(),
+            title: title.into_owned(),
+            description: description.into_owned(),
+            favicon_url: favicon_url.into_owned(),
+            logo_url: logo_url.into_owned(),
+            banner_url: banner_url.unwrap().into_owned(),
```
kespinola, line 86: What is a better why to handle this? The code defaults to saving the records with "" but can be none.

crates/rpc/src/rpc_models.rs
```diff
@@ -162,6 +164,7 @@ impl<'a> From<models::Storefront<'a>> for Storefront {
             favicon_url: favicon_url.into_owned(),
             logo_url: logo_url.into_owned(),
             updated_at: updated_at.map(Timestamp::from_utc),
+            banner_url: banner_url.unwrap().into_owned(),
```
ray-kast, line 167: Same as above

crates/graphql/src/schema.rs
ray-kast: Does GraphQL support `banner_url` being an `Option<String>`?  If so it would simplify the code to something like `banner_url.map(Cow::into_owned)`

crates/rpc/src/rpc_models.rs
ray-kast, line 167: Calling `.unwrap()` may panic

crates/indexer/src/bits/get_storefronts.rs
```diff
@@ -145,6 +148,7 @@ fn process_tags(
             description: Owned(description),
             favicon_url: Owned(favicon_url),
             logo_url: Owned(logo_url),
+            banner_url: Some(Owned(banner_url)),
```
ray-kast, line 151:
```suggestion
            banner_url: banner_url.map(Owned),
```

crates/indexer/src/bits/get_storefronts.rs
```diff
@@ -137,6 +137,9 @@ fn process_tags(
         let logo_url = tags
             .remove("holaplex:theme:logo:url")
             .unwrap_or_else(String::new);
+        let banner_url = tags
+            .remove("holaplex:theme:banner:url")
+            .unwrap_or_else(String::new);
```
ray-kast, line 142:
```suggestion
        let banner_url = tags.remove("holaplex:theme:banner:url");
```

ray-kast (changes requested):
I think there's a couple tweaks that can maybe be made for smoother Option code.

ray-kast (approved):
:tada:

PR #82, opened 2022-01-26
Access DB Pool from Context

kespinola:
### Change

Access the db_pool from the connection so its available more widely within graphql instead of on the query root.

### DIFF
--- a/crates/graphql/src/main.rs
+++ b/crates/graphql/src/main.rs
@@ -36,19 +36,17 @@ fn graphiql(uri: String) -> impl Fn() -> HttpResponse + Clone {
 }
 
 fn graphql(
-    db_pool: &Arc<Pool>,
+    db_pool: Arc<Pool>,
 ) -> impl Fn(
     web::Data<Arc<Schema>>,
     web::Json<GraphQLRequest>,
 ) -> Pin<Box<dyn Future<Output = Result<HttpResponse, Error>> + Send>>
 + Clone {
-    let db_pool = Arc::clone(db_pool);
-
     move |st, data| {
-        let db_connection = db_pool.get().unwrap();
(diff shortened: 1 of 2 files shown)


crates/graphql/src/main.rs
```diff
@@ -75,7 +73,7 @@ fn main() {
 
         let graphiql_uri = format!("http://{}", addr);
 
-        let schema = Arc::new(schema::create(db_pool.clone()));
+        let schema = Arc::new(schema::create());
```
kespinola, line 76: Schema prob doesn't need to be an arc now.

ray-kast: @kespinola that may be a juniper thing, although it'd be neat if we could drop the arc

crates/graphql/src/schema.rs
```diff
@@ -166,12 +164,16 @@ impl BatchFn<String, Option<NftDetail>> for NftDetailBatcher {
...
 
 impl AppContext {
-    pub fn new(db_connection: PooledConnection) -> AppContext {
+    pub fn new(db_pool: Arc<Pool>) -> AppContext {
         Self {
-            nft_detail_loader: Loader::new(NftDetailBatcher { db_connection }),
+            nft_detail_loader: Loader::new(NftDetailBatcher {
+                db_pool: db_pool.clone(),
+            }),
+            db_pool,
```
kespinola, line 176: Since arc is used clone each reference to db_pool besides the last within the block.

crates/graphql/src/main.rs
kespinola: Can try in a later PR.

ray-kast (approved):

PR #83, opened 2022-01-26
Query Creator Attribute Groups

kespinola:
### Goal
A client can query the set of attribute groups associated to a creator's nfts with count of each variant.
A client can filter the nfts root query by attributes.

```graphql
creator(address: "{address}") {
  attributeGroups {
    name
     variants {
         name
         count
     }
   }
}
```


crates/core/src/db/models.rs
```diff
@@ -263,8 +263,22 @@ pub struct File<'a> {
...
+pub struct MetadataAttributeWrite<'a> {
+    /// Metadata address
+    pub metadata_address: Cow<'a, str>,
+    /// Attribute name
+    pub name: Option<Cow<'a, str>>,
+    /// Attribute value
+    pub value: Option<Cow<'a, str>>,
+    /// Attribute trait type
+    pub trait_type: Option<Cow<'a, str>>,
+}
```
kespinola, line 277: When using generated fields within the db there needs to be a separate write vs read model.

kespinola: The write has insertable and query the Queryable derive.

In my reading on this the docs also suggest not mixing a struct as both queryable and insertable. Thoughts @ray-kast

crates/core/src/db/models.rs
```diff
@@ -273,6 +287,8 @@ pub struct Attribute<'a> {
     pub value: Option<Cow<'a, str>>,
     /// Attribute trait type
     pub trait_type: Option<Cow<'a, str>>,
+    /// Attribute generated id
+    pub id: Cow<'a, uuid::Uuid>,
```
kespinola, line 291: The model types can not be deisel types. Must always be rust type hence pulling in the uuid crate.

crates/graphql/src/schema.rs
```diff
@@ -5,13 +5,92 @@ use dataloader::{non_cached::Loader, BatchFn};
...
+                });
+            }
+
+            attribute_groups.push(Property {
+                name: attribute_group_name,
+                variants,
+            });
+        }
+
+        attribute_groups
```
kespinola, line 78: My ideal for this is to have database views that recalculated based on computed columns but this should get us solid traction. Can cache the queries with a ttl as the next improvement.

kespinola: I opted for commuting with rust vs making a bunch of select where count queries within sql.

crates/indexer/src/bits/metadata_uri.rs
```diff
@@ -119,11 +119,15 @@ enum ValueDataType {
...
-impl fmt::Display for ValueDataType {
+impl Display for ValueDataType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:?}", self)
+        match self {
+            Self::Number(n) => Display::fmt(n, f),
+            Self::String(s) => Display::fmt(s, f),
+        }
     }
 }
```
kespinola, line 129: Patch @ray-kast put together to fix the writing of the values to the db.

crates/graphql/src/schema.rs
```diff
@@ -5,13 +5,92 @@ use dataloader::{non_cached::Loader, BatchFn};
...
+    fn address(&self) -> String {
+        self.address.clone()
+    }
+
+    pub fn attribute_groups(&self, context: &AppContext) -> Vec<Property> {
+        let conn = context.db_pool.get().unwrap();
+
+        let metadatas: Vec<String> = metadata_creators::table
+            .select(metadata_creators::metadata_address)
+            .filter(metadata_creators::creator_address.eq(self.address.clone()))
```
kespinola, line 31: This is a nifty feature of graphql is using the context of the parent node in downstream lookups. very powerful.

crates/core/src/db/models.rs
ray-kast: Yeah, for structs where we have write-default fields separate would be good.  Diesel generally recommend making purpose-built models, but I merged them where possible to reduce boilerplate.

crates/graphql/src/schema.rs
ray-kast: Yeah, diesel only really supports collating rows outside of SQL.

ray-kast: @kespinola can i get a rundown on what this code is doing?

ray-kast (commented):
Looks good overall, although I'm curious if the double `for`-loop can be eliminated.

kespinola:
@ray-kast its taking all the attributes for the metadata associated to the creator and reducing them to a map that looks like this

```
{"Size": {"Small (S)": 14, "Extra Large (XL)": 4, "Medium (M)": 9}, "Type": {"Land": 44}, "Light Source": {"Yes": 28}, "Edition": {"Founders Edition (FE)": 11, "Founders Edition": 5}, "Dynamic Bones": {"Yes": 28}, "Height": {"50m ": 1, "25n": 1, "25m": 33, "50m": 8, "25": 1}, "Lip Sync": {"Yes": 28}, "MBs": {"100": 7, "200": 4, "75": 7, "600": 4, "250": 5}, "Custom Shaders": {"Yes": 28}, "Full Body Tracking": {"Yes": 28}, "Size m²": {"600": 35, "1500": 9}, "Finger Tracking": {"Yes": 28}, "Parcel Size": {"Medium (M)": 35, "Extra Large (XL)": 9}}
```

Then transforming into a list structure for the ui which is more usable but less efficient for doing the counting.

crates/graphql/src/schema.rs
kespinola: @ray-kast its taking all the attributes for the metadata associated to the creator and reducing them to a map that looks like this
```
{"Size": {"Small (S)": 14, "Extra Large (XL)": 4, "Medium (M)": 9}, "Type": {"Land": 44}, "Light Source": {"Yes": 28}, "Edition": {"Founders Edition (FE)": 11, "Founders Edition": 5}, "Dynamic Bones": {"Yes": 28}, "Height": {"50m ": 1, "25n": 1, "25m": 33, "50m": 8, "25": 1}, "Lip Sync": {"Yes": 28}, "MBs": {"100": 7, "200": 4, "75": 7, "600": 4, "250": 5}, "Custom Shaders": {"Yes": 28}, "Full Body Tracking": {"Yes": 28}, "Size m²": {"600": 35, "1500": 9}, "Finger Tracking": {"Yes": 28}, "Parcel Size": {"Medium (M)": 35, "Extra Large (XL)": 9}}
```
Then transforming into a list structure for the ui which is more usable but less efficient for doing the counting.

ray-kast, line 78: Doing something like the following would be more idiomatic Rust:
```
    let properties = metadata_attributes
        .into_iter()
        .fold(
            HashMap::new(),
            |mut groups,
             MetadataAttribute {
                 trait_type, value, ..
             }| {
                *groups
                    .entry(trait_type)
                    .or_insert_with(HashMap::new)
                    .entry(value)
                    .or_insert(0) += 1;

                groups
            },
        )
        .into_iter()
        .map(|(name, vars)| Property {
            name,
            variants: vars
                .into_iter()
                .map(|(name, count)| PropertyVariant { name, count })
                .collect(),
        })
        .collect::<Vec<_>>();
```

ray-kast (changes requested):

ray-kast (approved):
:+1: :smile:

PR #98, opened 2022-02-01
heroku indexer script

dandlezzz:


ray-kast (approved):
🤞

ray-kast:
@dandlezzz the script file is marked as executable, right?  github won’t tell me whether it is

dandlezzz:
> @dandlezzz the script file is marked as executable, right? github won’t tell me whether it is

CHMOD YES

PR #113, opened 2022-02-04
Adds graphql nft attributes support

echohtp:
Graphql Dataloader for nft attributes


crates/graphql/src/schema.rs
```diff
@@ -181,6 +181,50 @@ impl<'a> From<models::MetadataCreator<'a>> for NftCreator {
...
+    }
+}
+
+impl<'a> From<models::MetadataAttribute<'a>> for NftAttribute {
+    fn from(
+        models::MetadataAttribute {
+            metadata_address,
+            value,
+            trait_type,
+            ..
```
echohtp, line 217: I tried to bring in `id` but was unable to get the type `uuid:Uuid` to `use` correctly

crates/graphql/src/schema.rs
```diff
@@ -580,6 +624,36 @@ impl BatchFn<String, Vec<NftCreator>> for NftCreatorBatcher {
...
+        let rows: Vec<models::MetadataAttribute> = attributes::table
+            .filter(attributes::metadata_address.eq(any(addresses)))
+            .load(&conn)
+            .unwrap();
+
+        rows.into_iter().map(Into::into).map(|attr: NftAttribute| {
+            hash_map
+                .entry(attr.metadata_address.to_string())
+                .or_insert_with(Vec::new)
+                .push(attr);
```
kespinola, line 650: This should be a fold and need to collect.

crates/graphql/src/schema.rs
ray-kast: We should probably re-export it from the indexer_core crate to make that easier.  It’s randomly generated by postgres so it’s not a huge deal either way if we don’t get it.

crates/graphql/src/schema.rs
```diff
@@ -181,6 +181,50 @@ impl<'a> From<models::MetadataCreator<'a>> for NftCreator {
     }
 }
 
+#[derive(Debug, Clone)]
+struct NftAttribute {
+    metadata_address: String,
+    value: String,
+    trait_type: String,
+}
+
```
ray-kast, line 190: Should this be adorned with a juniper attribute?

crates/graphql/src/schema.rs
```diff
@@ -580,6 +624,36 @@ impl BatchFn<String, Vec<NftCreator>> for NftCreatorBatcher {
...
+        for address in addresses {
+            hash_map.insert(address.clone(), Vec::new());
+        }
+
+        let rows: Vec<models::MetadataAttribute> = attributes::table
+            .filter(attributes::metadata_address.eq(any(addresses)))
+            .load(&conn)
+            .unwrap();
+
+        rows.into_iter().map(Into::into).map(|attr: NftAttribute| {
```
ray-kast, line 646: The clippy warnings here hint at the underlying issue - you have to call a method that consumes this iterator, otherwise it never runs as Rust executes iterators lazily

ray-kast (dismissed):

crates/graphql/src/schema.rs
echohtp: Yep!

echohtp:
image

kespinola (approved):

PR #121, opened 2022-02-07
Index auction houses. (Rebase from abdul/auction-house)

ray-kast:


crates/core/migrations/2022-02-04-230210_create_auction_house_table/up.sql
```diff
@@ -0,0 +1,15 @@
...
+  fee_withdrawal_destination      varchar(48) not null,
+  authority                       varchar(48) not null,   
+  creator                         varchar(48) not null,  
+  bump                            smallint    not null,
+  treasury_bump                   smallint    not null, 
+  fee_payer_bump                  smallint    not null,
+  seller_fee_basis_points         smallint    not null,
+  requires_sign_off               boolean     not null,
+  can_change_sale_price           boolean     not null
+);
```
ray-kast, line 15: Do we have a spec for the column set we'll need for auction houses?

kespinola: This is a 1 for 1 of the auction house account stored in the program. all the bumps and seller_fee_basis_points are u8.

ray-kast: Do we need all of these columns, or would it be safe to remove some of them?

imabdulbasit: maybe we dont need bump columns

kespinola: I'd like to keep the bumps. Most of the auction house actions require the bump to be passed in. We can save some time and reduce complexity by having the bumps handly instead of doing find_progrma_address all the time.

kespinola (approved):

PR #123, opened 2022-02-08
Auction GraphQL retrieval scaffold

echohtp:
Doesn't compile, model still needs to be created.


kespinola (changes requested):
I already have the listing object. lets sync. we do need the query root.

ray-kast:
Can I close?

PR #128, opened 2022-02-14
nfts by wallet

imabdulbasit:
Query NFTs by wallet address

### DIFF
--- a/crates/graphql/src/schema.rs
+++ b/crates/graphql/src/schema.rs
@@ -8,15 +8,15 @@ use indexer_core::{
         tables::{
             attributes, auction_caches, auction_datas, auction_datas_ext, auction_houses, bids,
             listing_metadatas, metadata_creators, metadata_jsons, metadatas, store_config_jsons,
-            storefronts,
+            storefronts, token_accounts,
         },
         Pool,
     },
     prelude::*,
 };
 use juniper::{
-    EmptyMutation, EmptySubscription, FieldResult, GraphQLInputObject, GraphQLObject,
-    ParseScalarResult, ParseScalarValue, RootNode, Value,
+    graphql_value, EmptyMutation, EmptySubscription, FieldError, FieldResult, GraphQLInputObject,
+    GraphQLObject, ParseScalarResult, ParseScalarValue, RootNode, Value,
 };
 use reqwest::Client as HttpClient;
 use serde::Deserialize;
@@ -1056,13 +1056,20 @@ impl QueryRoot {
     fn nfts(
         &self,
         context: &AppContext,
-        #[graphql(description = "Filter on creator address")] creators: Vec<String>,
+        #[graphql(description = "Filter on owner address")] owners: Option<Vec<String>>,
+        #[graphql(description = "Filter on creator address")] creators: Option<Vec<String>>,
         #[graphql(description = "Filter on attributes")] attributes: Option<Vec<AttributeFilter>>,
-    ) -> Vec<Nft> {
-        let conn = context.db_pool.get().unwrap();
+    ) -> FieldResult<Vec<Nft>> {
+        if owners.is_none() && creators.is_none() {
+            return Err(FieldError::new(
+                "No filter provided! Please provide at least one of the filters",
+                graphql_value!({ "Filters": "owners: Vec<String>, creators: Vec<String>" }),
+            ));
+        }
+
+        let conn = context.db_pool.get().context("failed to connect to db")?;
 
         let query = metadatas::table.into_boxed();
-
         let query = attributes.unwrap_or_else(Vec::new).into_iter().fold(
             query,
             |acc, AttributeFilter { trait_type, values }| {
@@ -1078,11 +1085,22 @@ impl QueryRoot {
             },
         );
 
+        let mint_addresses_subquery = token_accounts::table
+            .select(token_accounts::mint_address)
+            .filter(
+                token_accounts::owner_address
+                    .eq(any(owners.unwrap_or_else(Vec::new)))
+                    .and(token_accounts::amount.eq(1)),
+            );
+        let metadata_addresses_subquery = metadata_creators::table
+            .select(metadata_creators::metadata_address)
+            .filter(metadata_creators::creator_address.eq(any(creators.unwrap_or_else(Vec::new))));
+
         let rows: Vec<models::Nft> = query
             .filter(
-                metadatas::address.eq(any(metadata_creators::table
-                    .select(metadata_creators::metadata_address)
-                    .filter(metadata_creators::creator_address.eq(any(creators))))),
+                metadatas::address
+                    .eq(any(metadata_addresses_subquery))
+                    .or(metadatas::mint_address.eq(any(mint_addresses_subquery))),
             )
             .inner_join(
                 metadata_jsons::table.on(metadatas::address.eq(metadata_jsons::metadata_address)),
@@ -1098,9 +1116,9 @@ impl QueryRoot {
             ))
             .order_by(metadatas::name.desc())
             .load(&conn)
-            .unwrap();
+            .context("failed to load nft")?;
 
-        rows.into_iter().map(Into::into).collect()
+        Ok(rows.into_iter().map(Into::into).collect())
     }
 
     fn wallet(


crates/graphql/src/schema.rs
```diff
@@ -961,6 +961,31 @@ impl QueryRoot {
 
         rows.pop().map(Into::into)
     }
+
+    fn nfts_by_wallet(&self, context: &AppContext, owner_address: String) -> Vec<Nft> {
```
kespinola, line 965: Ideally I'd like this to be a filter in the existing nft query. That maybe be challenging but give it a look and let me know. nfts(owners: []) should be additive to the existing creators filter

kespinola (changes requested):

crates/graphql/src/schema.rs
imabdulbasit: I have made creators and owners as optional parameters

imabdulbasit: But the nfts query returns all the nfts when no parameter is passed , is that okay or do you want to return empty results when no parameter is passed

crates/graphql/src/schema.rs
```diff
@@ -897,9 +897,13 @@ impl QueryRoot {
     fn nfts(
         &self,
         context: &AppContext,
-        #[graphql(description = "Filter on creator address")] creators: Vec<String>,
+        #[graphql(description = "Filter on owner address")] owners: Option<Vec<String>>,
+        #[graphql(description = "Filter on creator address")] creators: Option<Vec<String>>,
         #[graphql(description = "Filter on attributes")] attributes: Option<Vec<AttributeFilter>>,
     ) -> Vec<Nft> {
+        if owners.is_none() && creators.is_none() && attributes.is_none() {
+            return Vec::new();
```
kespinola, line 905: Lets return a graphql api error here instead. I haven't looked that up yet.

kespinola (changes requested):
Looks great. Just the change to a graphql api error. something like one of the following filters is required...

kespinola:
I'll get the api error change in now.

crates/graphql/src/schema.rs
imabdulbasit: added

imabdulbasit: image

crates/graphql/src/schema.rs
```diff
@@ -919,12 +926,20 @@ impl QueryRoot {
...
+                    .eq(any(owners.unwrap_or_else(Vec::new)))
+                    .and(token_accounts::amount.eq(1)),
+            );
+        let metadata_addresses_subquery = metadata_creators::table
+            .select(metadata_creators::metadata_address)
+            .filter(metadata_creators::creator_address.eq(any(creators.unwrap_or_else(Vec::new))));
+
+        let rows: Vec<models::Nft> = query
+            .filter(metadatas::address.eq(any(metadata_addresses_subquery)))
+            .or_filter(metadatas::mint_address.eq(any(mint_addresses_subquery)))
```
kespinola, line 942: Can some folks audit me on this? I switched 2 subqueries 1 for finding metadata addresses associated to the creators passed in and another for mint_address for owners passed in.

Then use a WHERE address IN (creators_metadata_addressses) OR mint_address IN (mint_addresses)

This query should return metadata that match the attribute filters and either the NFT is created by one of the creators or owned by one of the owners.

kespinola (approved):

crates/graphql/src/schema.rs
imabdulbasit: should work

crates/graphql/src/schema.rs
```diff
@@ -919,11 +926,22 @@ impl QueryRoot {
             },
         );
 
+        let mint_addresses_subquery = token_accounts::table
+            .select(token_accounts::mint_address)
+            .filter(
+                token_accounts::owner_address
+                    .eq(any(owners.unwrap_or_else(Vec::new)))
+                    .and(token_accounts::amount.eq(1)),
```
kespinola, line 934: A great index candidate [mint_address, amount === 1]

crates/graphql/src/schema.rs
```diff
@@ -919,11 +926,22 @@ impl QueryRoot {
...
+            .filter(metadata_creators::creator_address.eq(any(creators.unwrap_or_else(Vec::new))));
+
         let rows: Vec<models::Nft> = query
             .filter(
-                metadatas::address.eq(any(metadata_creators::table
-                    .select(metadata_creators::metadata_address)
-                    .filter(metadata_creators::creator_address.eq(any(creators))))),
+                metadatas::address
+                    .eq(any(metadata_addresses_subquery))
+                    .or(metadatas::mint_address.eq(any(mint_addresses_subquery))),
```
kespinola, line 944: Another index for mint_address on metadatas table.

crates/graphql/src/schema.rs
```diff
@@ -919,11 +926,22 @@ impl QueryRoot {
...
+        let mint_addresses_subquery = token_accounts::table
+            .select(token_accounts::mint_address)
+            .filter(
+                token_accounts::owner_address
+                    .eq(any(owners.unwrap_or_else(Vec::new)))
+                    .and(token_accounts::amount.eq(1)),
+            );
+        let metadata_addresses_subquery = metadata_creators::table
+            .select(metadata_creators::metadata_address)
+            .filter(metadata_creators::creator_address.eq(any(creators.unwrap_or_else(Vec::new))));
```
kespinola, line 938: Index on creator for metadata_creators.

kespinola (commented):
notes on indexes to add.

crates/graphql/src/schema.rs
```diff
@@ -897,13 +897,20 @@ impl QueryRoot {
...
         context: &AppContext,
-        #[graphql(description = "Filter on creator address")] creators: Vec<String>,
+        #[graphql(description = "Filter on owner address")] owners: Option<Vec<String>>,
+        #[graphql(description = "Filter on creator address")] creators: Option<Vec<String>>,
         #[graphql(description = "Filter on attributes")] attributes: Option<Vec<AttributeFilter>>,
-    ) -> Vec<Nft> {
+    ) -> FieldResult<Vec<Nft>> {
+        if owners.is_none() && creators.is_none() {
+            return Err(FieldError::new(
+                "No filter provided! Please provide atleast one of the filter",
```
ray-kast, line 906:
```suggestion
                "No filter provided! Please provide at least one of the filters",
```

ray-kast (changes requested):
Looks good

crates/graphql/src/schema.rs
```diff
@@ -941,7 +959,7 @@ impl QueryRoot {
             .load(&conn)
             .unwrap();
 
-        rows.into_iter().map(Into::into).collect()
+        Ok(rows.into_iter().map(Into::into).collect())
```
ray-kast, line 962: This function calls `.unwrap()` in several spots, so if the return type is to be changed to Result the panics should also be removed.

ray-kast (changes requested):

crates/graphql/src/schema.rs
ray-kast: @imabdulbasit i see you resolved this, did the fix get committed?

crates/graphql/src/schema.rs
imabdulbasit: this index is already added added

crates/graphql/src/schema.rs
imabdulbasit: I pushed a commit

crates/graphql/src/schema.rs
imabdulbasit: yeah this needs to be changed in whole schema.rs

ray-kast: Right.  For now I'd suggest just removing the FieldResult if possible and unwrapping.

crates/graphql/src/schema.rs
ray-kast: Changeset doesn't show it was fixed, are you sure?

imabdulbasit: https://github.com/holaplex/solana-indexer/pull/128/commits/be6664be3d22cbf6f4b61731ef4b7d3780118467

ray-kast: :+1: aight idk what happened on github's end lol

crates/graphql/src/schema.rs
imabdulbasit: I have changed FieldResult to Result<Vec<Nft>,FieldError>

ray-kast: I think having FieldResult was fine, as long as all unwraps are changed to the `?` operator.

ray-kast:
@imabdulbasit you'll need to rebase onto dev

ray-kast (approved):

PR #141, opened 2022-02-16
remove debug statements, outside of ah

dandlezzz:
want to debug production a bit. gonna cut down on the noise.


ray-kast:
Production won’t actually print debug! statements

PR #154, opened 2022-02-18
Only Index Tokens with amount 0 or 1

kespinola:
### Changes
- Only index tokens related to NFTs where the amount is always 0 or 1.

### DIFF
--- a/crates/indexer/src/accountsdb/accounts/token.rs
+++ b/crates/indexer/src/accountsdb/accounts/token.rs
@@ -1,95 +1,14 @@
 use chrono::offset::Local;
 use indexer_core::{
-    db::{
-        insert_into,
-        models::{TokenAccount as TokenAccountModel, TokenTransfer},
-        select,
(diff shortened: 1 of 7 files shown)


crates/indexer/src/accountsdb/accounts/token.rs
```diff
@@ -1,95 +1,14 @@
...
-
-    Ok(())
-}
 pub async fn process(client: &Client, key: Pubkey, token_account: TokenAccount) -> Result<()> {
-    if token_account.mint == sol() {
-        return Ok(());
-    }
-    if token_account.mint == kin() {
-        return Ok(());
-    }
```
ray-kast, line 72: I think we could either keep this, or if we remove it we should get rid of the definitions for `sol` and `kin` in pubkeys.rs

ray-kast (changes requested):

ray-kast (approved):

PR #162, opened 2022-02-21
deploy

dandlezzz:


crates/accountsdb-rabbitmq/Cargo.toml
```diff
@@ -30,6 +30,8 @@ solana-accountsdb-plugin-interface = "1.9.4"
 solana-logger = "1.9.4"
 solana-program = "1.9.4"
 solana-sdk = "1.9.4"
+spl-token = "3.2.0"
+reqwest = "*"
```
ray-kast, line 34: We should make this an explicit version, depending on * can break the build without warning.

crates/accountsdb-rabbitmq/src/plugin.rs
```diff
@@ -46,6 +70,21 @@ impl AccountsDbPlugin for AccountsDbPluginRabbitMq {
...
+            .expect("couldn't fetch token list");
+        let mut body = String::new();
+        res.read_to_string(&mut body)?;
+
+        let json: TokenList = serde_json::from_str(&body).expect("file should be proper JSON");
+
+        let mut token_addresses: HashMap<String, bool> = HashMap::new();
+        for token_data in json.tokens {
+            let address = token_data.address;
+            token_addresses.insert(address, true);
```
ray-kast, line 83: Should use a HashSet for this instead.

ray-kast (changes requested):
Token registry PR needs some fixups first.

ray-kast (dismissed):

ray-kast (approved):

PR #166, opened 2022-02-21
Make the accountsdb plugin sensitive to `is_startup`.

ray-kast:


dandlezzz (approved):
Looks good. Thanks for the refactorings.

PR #167, opened 2022-02-22
Oops.

ray-kast:


kespinola (approved):

crates/accountsdb-rabbitmq/src/plugin.rs
```diff
@@ -58,12 +58,11 @@ struct TokenList {
 impl AccountsDbPluginRabbitMq {
     const TOKEN_REG_URL: &'static str = "https://raw.githubusercontent.com/solana-labs/token-list/main/src/tokens/solana.tokenlist.json";
 
-    async fn load_token_reg() -> Result<HashSet<Pubkey>> {
-        let res: TokenList = reqwest::get(Self::TOKEN_REG_URL)
-            .await
+    fn load_token_reg() -> Result<HashSet<Pubkey>> {
+        // We use `smol` as an executor, and reqwest's async backend doesn't like that
```
kespinola, line 62: Up for dropping the comment on this PR then can see it in the blame.

PR #171, opened 2022-02-22
Refactor GraphQL + Add bid indexing

ray-kast:


crates/graphql/src/schema/strings.rs
```diff
@@ -0,0 +1,55 @@
+macro_rules! strings {
```
kespinola, line 1: Talked in huddle but my preference is to fork the data loader and add result support vs building a macro. Like people to read vanilla docs and be able to contribute without having to lookup custom macros.

crates/graphql/src/schema/util.rs
```diff
@@ -0,0 +1,30 @@
+use juniper::{ParseScalarResult, ParseScalarValue, Value};
```
kespinola, line 1: can this be in scalars module. Likely to add more. not a util per say.

kespinola (changes requested):
Advocate for dropping the `/schema/` subdir. Really love this. Ty.

crates/graphql/src/schema/util.rs
kespinola: still think we should track this in the scalars directory.

crates/graphql/src/schema/strings.rs
ray-kast: This one's actually not related to data loaders.  The macro there is just a bunch of repetitive struct and trait declarations for a bunch of near-identical structs, so it's only called once to declare all of them.  The only time it's called is [here](https://github.com/holaplex/solana-indexer/pull/171/files/59a8c32b05f2adcb54600fd6b557bd540cb6cde8#diff-2e1bcce04320328d2428e0eb8c7b392180b7264a80ce99a157666f9ecbfe9b82R50-R55) to declare a bunch of structs in one go.

crates/graphql/src/schema/util.rs
ray-kast: Ah, missed this comment lol.

crates/graphql/src/schema/context.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub nft_attributes_loader: Loader<MetadataAddress, Vec<NftAttribute>>,
+    pub nft_creators_loader: Loader<MetadataAddress, Vec<NftCreator>>,
+    pub storefront_loader: Loader<StorefrontAddress, Option<Storefront>>,
+}
+
+impl juniper::Context for AppContext {}
+
+impl AppContext {
+    pub fn new(db_pool: Arc<Pool>, twitter_bearer_token: Arc<String>) -> AppContext {
+        let batcher = Batcher::new(db_pool.clone());
```
kespinola, line 31: nice.

crates/graphql/src/schema/dataloaders/batcher.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    pub fn diesel_batch<
+        K: Clone + Eq + std::hash::Hash,
+        KR: Into<K>,
+        V: BatchExtend,
+        M,
+        Q: LoadQuery<Connection, M>,
+        E: Into<indexer_core::error::Error>,
+    >(
```
kespinola, line 66: Chatted in the office but like to ship the batcher trait without `diesel_batch` to start. Lets go with some code duplication for now and drop the result until we update data batcher to support it.

crates/graphql/src/schema/objects/listing.rs
```diff
@@ -0,0 +1,139 @@
...
+    String,                // store_address
+    Option<NaiveDateTime>, // ends_at
+    Option<i32>,           // gap_time
+    Option<NaiveDateTime>, // last_bid_time
+);
+
+impl Listing {
+    pub fn address((address, ..): &ListingRow) -> String {
+        address.clone()
+    }
```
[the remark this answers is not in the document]
ray-kast: Used to pick out the address string without constructing a listing, since Listing::new isn't guaranteed to succeed.

kespinola (approved):

crates/graphql/src/schema/dataloaders/listing.rs
```diff
@@ -0,0 +1,97 @@
...
+                    metadatas::mint_address,
+                    metadatas::primary_sale_happened,
+                    metadata_jsons::description,
+                    metadata_jsons::image,
+                ),
+            ))
+            .load(&conn)
+            .context("Failed to load listing NFTs")?;
+
+        Ok(rows.into_iter().map(|(k, v)| (k, v.try_into())).batch(keys))
```
kespinola, line 95: If all the keys dont have a record when the loader tries to resolve it won't be present. Is that case handled bc it results in a runtime error.

ray-kast: yeah, `.batch` takes `keys` as a parameter so that it can populate the initial map.

kespinola:
Adds more queries likely affect refactor.

https://github.com/holaplex/solana-indexer/pull/186

PR #172, opened 2022-02-23
reduce filter options, eliminate subqueries

dandlezzz:


crates/graphql/src/schema.rs
```diff
@@ -1066,6 +1066,12 @@ impl QueryRoot {
                 graphql_value!({ "Filters": "owners: Vec<String>, creators: Vec<String>" }),
             ));
         }
+        if owners.is_some() && creators.is_some() {
+            return Err(FieldError::new(
+                "Please pass either owner or creator, not both",
```
kespinola, line 1071:
```suggestion
                "Please pass either owners or creators, not both",
```

kespinola (approved):

PR #174, opened 2022-02-23
Abdul/active-offers

imabdulbasit:
```
nft(address : " ") {
  offers {
     wallet
     price
     tradeStateAddress
     tradeStateBump
     activatedAt
  }
}
```

### DIFF
--- a/crates/graphql/src/schema.rs
+++ b/crates/graphql/src/schema.rs
@@ -7,8 +7,8 @@ use indexer_core::{
         models,
         tables::{
             attributes, auction_caches, auction_datas, auction_datas_ext, auction_houses, bids,
-            listing_metadatas, metadata_creators, metadata_jsons, metadatas, store_config_jsons,
-            storefronts, token_accounts,
+            listing_metadatas, metadata_creators, metadata_jsons, metadatas, public_bids,
+            store_config_jsons, storefronts, token_accounts,
         },
         Pool,
     },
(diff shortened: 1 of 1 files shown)


crates/graphql/src/schema.rs
```diff
@@ -680,6 +684,75 @@ impl<'a> From<models::AuctionHouse<'a>> for AuctionHouse {
...
+        self.token_mint.clone()
+    }
+    pub fn price(&self) -> UnsignedInt8 {
+        self.price
+    }
+    pub fn trade_state_bump(&self) -> i32 {
+        self.trade_state_bump
+    }
+
+    pub fn activated_at(&self) -> Option<UnsignedInt8> {
```
kespinola, line 720: Can we convert these to datetime utc?

Same with closed_at

kespinola: May be best to do this conversion when saving to the db. use datatime col type in postgres.


Then we can add a custom scaler for DateTime if not supported natively by juniper.

crates/graphql/src/schema.rs
```diff
@@ -27,18 +27,18 @@ struct Creator {
 }
 
 #[derive(Debug, Clone, Copy)]
-pub struct Lamports(u64);
+pub struct UnsignedInt8(u64);
 
-#[juniper::graphql_scalar(description = "Lamports")]
-impl<S> GraphQLScalar for Lamports
+#[juniper::graphql_scalar(description = "UnsignedInt8")]
+impl<S> GraphQLScalar for UnsignedInt8
```
kespinola, line 33: If we switch activated_at to date time scaler can this stay as a lamports scalar.

imabdulbasit: yes

PR #179, opened 2022-02-24
Abdul/ah-receipts

imabdulbasit:
- Listing, Purchase and PublicBid indexing
- Query for active NFT offers
- Query for NFTs up for Sale


crates/graphql/src/schema.rs
```diff
@@ -690,6 +693,75 @@ impl<'a> From<models::AuctionHouse<'a>> for AuctionHouse {
...
 
+#[derive(Debug, Clone)]
+struct PublicBid {
+    pub address: String,
+    pub trade_state: String,
+    pub wallet: String,
+    pub token_mint: String,
+    pub price: Lamports,
+    pub trade_state_bump: i32,
+    pub activated_at: Option<DateTime<Utc>>,
```
imabdulbasit, line 704: I can change this to NaiveDateTime but if we want date and time then this should be okay

crates/graphql/src/schema.rs
```diff
@@ -497,6 +497,18 @@ impl Nft {
     pub async fn owner(&self, ctx: &AppContext) -> Option<NftOwner> {
         ctx.nft_owner_loader.load(self.mint_address.clone()).await
     }
+    pub async fn offers(&self, ctx: &AppContext) -> Vec<PublicBid> {
+        ctx.public_bid_loader.load(self.mint_address.clone()).await
+    }
+
+    pub async fn listing(&self, ctx: &AppContext) -> Vec<AuctionHouseListing> {
```
kespinola, line 504: This one is interesting because there can only be 1 listing for the auction house but the same nft can be listed on multiple auction houses.

We can do this 2 ways:

1. Make listing take an argument of ah so we can return the single potential listing for the nft.
2. Make this field `listings` and return vec[listings] like you are now. 

I advocate for the 2nd approach and the client filters this list based on the ah scope of the marketplace.

kespinola: The other item is the use of 2 loaders. The two dedicated loaders is good if most clients will query the nft owner in the same requests. Though this is true for our case I still advocate for the auction_house_listing loader to handle the token_account lookup.

The loader should take mint addresses query token_accounts where mint_address = nft.mint_address AND amount = 1. Then inner join on listings with token_account.owner_address = listing.seller AND listing.activated_at IS NOT NUL AND listing.canceled_at IS NUL.

crates/graphql/src/schema.rs
```diff
@@ -690,6 +702,136 @@ impl<'a> From<models::AuctionHouse<'a>> for AuctionHouse {
...
+        }: models::PublicBid,
+    ) -> Result<Self, Self::Error> {
+        Ok(Self {
+            address: address.into_owned(),
+            trade_state: trade_state.into_owned(),
+            wallet: wallet.into_owned(),
+            token_mint: token_mint.into_owned(),
+            price: price.try_into()?,
+            trade_state_bump: trade_state_bump.into(),
+            activated_at: activated_at.map(|x| DateTime::<Utc>::from_utc(x, Utc)),
```
kespinola, line 769: nice. ty.

kespinola (changes requested):
looking really good. just item to adjust the auction_house_listings loader.

crates/graphql/src/schema.rs
kespinola:
```graphql
nfts {
  listings {
       price
          tradeState
           auctionHouse
      }
  }
}
```

```sql
SELECT * LISTINGS INNER JOIN listing.seller = token_accounts.owner 
\ WHERE listing.token_mint = nft.mint_address AND token_account.amount = 1;
```


Discussed in standup this is the query to use.

crates/graphql/src/schema.rs
```diff
@@ -999,6 +1153,89 @@ impl BatchFn<String, Vec<AuctionHouse>> for AuctionHouseBatcher {
...
+        let mut hash_map = HashMap::new();
+
+        for address in metadata_addresses {
+            hash_map.insert(address.clone(), Vec::new());
+        }
+
+        let rows: Vec<models::BidReceipt> = bid_receipts::table
+            .inner_join(metadatas::table.on(metadatas::address.eq(bid_receipts::metadata)))
+            .inner_join(
+                token_accounts::table.on(token_accounts::mint_address.eq(metadatas::mint_address)),
```
imabdulbasit, line 1173: maybe we don't need to check token amount and remove this extra join.

ray-kast (dismissed):
Conflicts with current `dev`

crates/graphql/src/schema.rs
kespinola: yeah. we are going to use ah public bids which are associated to the nft and not the owner. If we did private sales then only want to return if the bid is only valid if its owned by the person it was set for.

kespinola (approved):

kespinola (approved):

PR #182, opened 2022-02-24
ugh we need schemas

dandlezzz:
Frustrating! Hard to enforce schema on something like this. Left debug statements in there intentionally so that we can resolve this sort of thing more quickly if it were to happen again.

Proof of market.

image

### DIFF
--- a/crates/indexer/src/accountsdb/programs/metadata.rs
+++ b/crates/indexer/src/accountsdb/programs/metadata.rs
@@ -66,6 +66,7 @@ pub(crate) async fn process(client: &Client, update: AccountUpdate) -> Result<()
         MASTER_EDITION_V1 => process_master_edition_v1(client, update).await,
         MASTER_EDITION_V2 => process_master_edition_v2(client, update).await,
         b => {
+            trace!("Unhandled metadata key byte {:02x}", b);
 
             Ok(())
         },
--- a/crates/indexer/src/accountsdb/programs/metadata.rs
+++ b/crates/indexer/src/accountsdb/programs/metadata.rs
@@ -66,7 +66,6 @@ pub(crate) async fn process(client: &Client, update: AccountUpdate) -> Result<()
         MASTER_EDITION_V1 => process_master_edition_v1(client, update).await,
         MASTER_EDITION_V2 => process_master_edition_v2(client, update).await,
         b => {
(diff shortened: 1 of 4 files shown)


kespinola (approved):

crates/indexer/src/accountsdb/programs/metadata.rs
```diff
@@ -66,7 +66,6 @@ pub(crate) async fn process(client: &Client, update: AccountUpdate) -> Result<()
         MASTER_EDITION_V1 => process_master_edition_v1(client, update).await,
         MASTER_EDITION_V2 => process_master_edition_v2(client, update).await,
         b => {
-            debug!("Unhandled metadata key byte {:02x}", b);
```
ray-kast, line 69: Change this to `trace!` rather than deleting.

crates/indexer/src/accountsdb/programs/metaplex.rs
```diff
@@ -88,7 +88,6 @@ pub(crate) async fn process(client: &Client, update: AccountUpdate) -> Result<()
         MPL_STORE_CONFIG => process_store_config(client, update).await,
         AUCTION_CACHE => process_auction_cache(client, update).await,
         b => {
-            debug!("Unhandled metadata key byte {:02x}", b);
```
ray-kast, line 91: Same here.

crates/indexer/src/http/store_config.rs
```diff
@@ -24,11 +24,23 @@ pub struct Address {
     pub auctionHouse: String,
```
ray-kast, line 24:
```suggestion
    pub auction_house: String,
```

Add `#[serde(rename_all = "camelCase")]` in between line 21 and 22.

crates/indexer/src/http/store_config.rs
```diff
@@ -24,11 +24,23 @@ pub struct Address {
...
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Upload {
+    pub url: String,
+    pub name: String,
+    pub r#type: String,
```
ray-kast, line 31:
```suggestion
    #[serde(rename = "type")]
    pub ty: String,
```

ray-kast (changes requested):

ray-kast (approved):

PR #185, opened 2022-02-25
remove breaking toLowerCase on trait_type

dandlezzz:
Hopefully there wasn't a reason behind this. It breaks most filter sets. 

If we did have a reason for this, one route might be to add a to_lowercase to all incoming filters. We'd have to change the query and utilize a LOWER. Both could work. Not sure which is better at the moment. Let's talk.


kespinola:
done here.

https://github.com/holaplex/solana-indexer/pull/186

PR #187, opened 2022-03-01
abdul/version-update

imabdulbasit:
anchor-lang 
mpl-auction-house

### DIFF
--- a/crates/indexer/Cargo.toml
+++ b/crates/indexer/Cargo.toml
@@ -42,14 +42,14 @@ tokio-amqp = "1.0.1"
 
 # Solana & Metaplex
 anchor-lang = "~0.20.1"
-metaplex = "0.0.1"
-metaplex-auction = "0.0.1"
-metaplex-token-metadata = "0.0.1"
-metaplex-token-vault = "0.0.1"
-solana-client = { version = "1.8.3", optional = true }
-solana-program = { version = "1.8.3", optional = true }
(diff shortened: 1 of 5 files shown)


crates/indexer/Cargo.toml
```diff
@@ -41,11 +41,12 @@ tokio = { version = "1.13.0", features = ["rt"] }
 tokio-amqp = "1.0.1"
 
 # Solana & Metaplex
-anchor-lang = "0.19.0"
+anchor-lang = "0.20.0"
 metaplex = "0.0.1"
 metaplex-auction = "0.0.1"
-metaplex-token-metadata = "0.0.1"
+mpl-token-metadata = { version="~1.2.4", features = [ "no-entrypoint" ] }
```
ray-kast, line 47:
```suggestion
mpl-token-metadata = { version="~1.2.4", features = ["no-entrypoint"] }
```

Should we add `no-entrypoint` to the other MPL imports?

ray-kast (changes requested):

crates/indexer/Cargo.toml
imabdulbasit: we should, metaplex uses it

ray-kast: @imabdulbasit do you wanna go ahead and fold that into this PR? seems like a good time to do it

ray-kast (approved):

PR #191, opened 2022-03-02
creators indexing

imabdulbasit:


kespinola:
store config json now has `address.store` which is the store pda address.

PR #193, opened 2022-03-03
Abdul/store-creators

imabdulbasit:
- creators indexing
- data loader


kespinola:
I added the store address to the json not the store_config let me add that too. It will be under address.storeConfig

kespinola:
https://github.com/holaplex/holaplex/pull/248

Ok address.storeConfig available as well.

crates/core/migrations/2022-03-02-195130_store_config_creators_table/up.sql
```diff
@@ -0,0 +1,8 @@
+create table store_creators (
+    store_config_address varchar(48),
+    creator_address      varchar(48),
+    PRIMARY KEY(store_config_address,creator_address)
```
ray-kast, line 4:
```suggestion
    primary key (store_config_address, creator_address)
```

crates/core/migrations/2022-03-02-220455_add_store_addres_col_to_store_config_jsons_table/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table store_config_jsons
+add column store_address varchar(48) default ' ' not null;
```
ray-kast, line 2:
```suggestion
add column store_address varchar(48) default ' ' not null;
```
This will likely cause deserialization errors in the live db, can we make this nullable instead?

crates/core/src/db/models.rs
```diff
@@ -10,7 +10,8 @@ use super::schema::{
     attributes, auction_caches, auction_datas, auction_datas_ext, auction_houses, bid_receipts,
     bids, editions, files, listing_metadatas, listing_receipts, master_editions,
     metadata_collections, metadata_creators, metadata_jsons, metadatas, purchase_receipts,
-    store_config_jsons, store_configs, storefronts, stores, token_accounts, whitelisted_creators,
+    store_config_jsons, store_configs, store_creators, store_denylist, storefronts, stores,
```
ray-kast, line 13:
```suggestion
    store_config_jsons, store_configs, store_creators, storefronts, stores,
```

crates/graphql/src/schema/objects/marketplace.rs
```diff
@@ -63,7 +72,9 @@ impl Marketplace {
     pub fn owner_address(&self) -> &str {
         &self.owner_address
     }
-
+    pub fn store_address(&self) -> &str {
+        &self.store_address
+    }
```
ray-kast, line 77:
```suggestion

    pub fn store_address(&self) -> &str {
        &self.store_address
    }

```

ray-kast (changes requested):

kespinola (approved):

crates/graphql/src/schema/objects/store_creator.rs
```diff
@@ -0,0 +1,32 @@
+use super::prelude::*;
+
+#[derive(Debug, Clone)]
+pub struct StoreCreator {
+    pub store_config_address: String,
+    pub creator_address: String,
+}
+
+#[graphql_object(Context = AppContext)]
```
ray-kast, line 9: This should just be a `derive(GraphQLObject)`

ray-kast (changes requested):

crates/graphql/src/schema/objects/store_creator.rs
kespinola: true. maybe but we can switch declaration at that time. sorry should have caught that too.

imabdulbasit: oh yes , I did add one more field to test smth and then removed

ray-kast: maybe juniper will have fixed the api for this by that time :crossed_fingers: i missed it as well lol

ray-kast (approved):

PR #211, opened 2022-03-08
Improve handling of indexer jobs.

ray-kast:
Tasks for this PR:
- [x] Add support for parallel job processing on the indexer
- [x] Add support for retrying failed indexing jobs
- [x] Add exponential backoff
- [x] BONUS: make the binding address of HTTP servers configurable
- [x] Verify job parallelism functions correctly
- [x] Verify exponential backoff and dead-lettering function correctly


crates/indexer/src/util.rs
```diff
@@ -12,33 +12,63 @@ use {
...
+    mut acct: Account,
+    f: impl Send + FnOnce(AccountInfo<'_>) -> T + 'static,
+) -> Result<T> {
+    // NOTE: this is here because metaplex_auction makes the brilliant choice to
+    //       only provide parsing via AccountInfo, which further makes the
+    //       brilliant choice to store lamports as Rc<RefCell<&mut u64>>, which
+    //       does not implement Send and therefore cannot ever be used in
+    //       multithreaded code.  Thus the only way to access it is to spawn a
+    //       dedicated thread in Tokio's thread pool to run all code that needs
+    //       to use the account info and prevent the need for a Send impl.
```
kespinola, line 28:
```suggestion
    // NOTE: this is here because metaplex_auction
    //       only provide parsing via AccountInfo, and also stores lamports as Rc<RefCell<&mut u64>>, which
    //       does not implement Send and therefore cannot ever be used in
    //       multithreaded code.  Thus the only way to access it is to spawn a
    //       dedicated thread in Tokio's thread pool to run all code that needs
    //       to use the account info and prevent the need for a Send impl.
```

crates/indexer/src/bin/metaplex-indexer-accountsdb.rs
```diff
@@ -48,29 +49,18 @@ fn main() {
...
-                    Err(e) => error!("Failed to process message: {:?}", e),
-                }
-            }
-
-            warn!("AMQP server hung up!");
-
-            Ok(())
+            metaplex_indexer::amqp_consume(&params, consumer, move |m| {
+                let client = client.clone();
+                async move { metaplex_indexer::accountsdb::process_message(m, &*client).await }
```
kespinola, line 61: can you take us over this?

ray-kast:
Closes #238.

imabdulbasit (approved):

dandlezzz (approved):
approved.

PR #219, opened 2022-03-10
returning assets url in graphql. Moving AssetIdentifier logic to indexer-core

mpwsh:
- moved `AssetIdentifier` from `indexer/http/metadata_json` to `indexer-core/assets.rs`
- added function to return detected service and Cid in `AssetsIdentifier`
- implemented transforming url to `assets.holaplex.com` in graphql

### DIFF
--- a/crates/graphql/src/schema/objects/nft.rs
+++ b/crates/graphql/src/schema/objects/nft.rs
@@ -164,13 +164,13 @@ impl Nft {
     pub fn image(&self) -> String {
         let assets_cdn = "https://assets.holaplex.com";
         let asset = AssetIdentifier::new(&Url::parse(&self.image).unwrap());
-        if asset.arweave.is_some() {
+        if asset.arweave.is_some() && asset.ipfs.is_none() {
             format!(
                 "{}/arweave/{}",
                 assets_cdn,
                 Base64Display::with_config(&asset.arweave.unwrap().0, base64::URL_SAFE_NO_PAD)
             )
-        } else if asset.ipfs.is_some() {
+        } else if asset.ipfs.is_some() && asset.arweave.is_none() {
             format!("{}/ipfs/{}", assets_cdn, asset.ipfs.unwrap())
         } else {
             String::from(&self.image)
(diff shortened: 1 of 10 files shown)


ray-kast (changes requested):
I would prefer the logic for assembling a CDN link to be handled by the GraphQL server rather than the `get_cid_with_svc` function.

crates/graphql/src/schema/query_root.rs
```diff
@@ -177,7 +176,14 @@ impl QueryRoot {
             .load(&conn)
             .context("Failed to load metadata")?;
 
-        Ok(rows.pop().map(Into::into))
+        Ok(rows.pop().map(|mut n| {
```
kespinola, line 179: Its cleaner do this swap in the image method of the nft object instead of in the dataloader.

kespinola (changes requested):

crates/graphql/src/schema/query_root.rs
mpwsh: Added proposed changes in image method

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -158,8 +161,20 @@ impl Nft {
         &self.description
     }
 
-    pub fn image(&self) -> &str {
-        &self.image
+    pub fn image(&self) -> String {
+        let assets_cdn = "https://assets.holaplex.com";
+        let asset = AssetIdentifier::new(&Url::parse(&self.image).unwrap());
```
ray-kast, line 166: Should check if both fields are Some.  It is technically not impossible that a link is ambiguous, in which case I think returning the raw URL is not out of the question.

ray-kast (changes requested):

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -158,8 +161,20 @@ impl Nft {
         &self.description
     }
 
-    pub fn image(&self) -> &str {
-        &self.image
+    pub fn image(&self) -> String {
+        let assets_cdn = "https://assets.holaplex.com";
```
kespinola, line 165: Configure the asset host in later PR.

kespinola (approved):

crates/graphql/src/schema/objects/nft.rs
ray-kast: https://github.com/holaplex/solana-indexer/issues/227

ray-kast (approved):

PR #223, opened 2022-03-11
switch to inner join for receipts

dandlezzz:
In the future as we develop more and more complex queries, we have to remember to test with larger datasets. The dark magic of the postgres query planner operates much different when it is examining a set of 3 or 30,000. 

When querying on this creator `8CWKUvr9KDqFZww2q3xurag3FdjqmAUEVT97hX3ZBYz9` , this query never returns. I think its because the inner joins have so many rows that adding a row for every receipt record causes the final table to be far too large. 

Do we need multiple receipts per NFT in this query? It qualifies as listed with just 1 that matches the filters correct?

### DIFF
--- a/crates/core/src/db/queries/metadatas.rs
+++ b/crates/core/src/db/queries/metadatas.rs
@@ -63,9 +63,7 @@ pub fn list(
         .inner_join(
             token_accounts::table.on(metadatas::mint_address.eq(token_accounts::mint_address)),
         )
-        .left_outer_join(
-            listing_receipts::table.on(metadatas::address.eq(listing_receipts::metadata)),
-        )
+        .inner_join(listing_receipts::table.on(metadatas::address.eq(listing_receipts::metadata)))
         .into_boxed();
 
     if let Some(attributes) = attributes {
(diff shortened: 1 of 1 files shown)


crates/core/src/db/queries/metadatas.rs
```diff
@@ -63,9 +63,7 @@ pub fn list(
         .inner_join(
             token_accounts::table.on(metadatas::mint_address.eq(token_accounts::mint_address)),
         )
-        .left_outer_join(
-            listing_receipts::table.on(metadatas::address.eq(listing_receipts::metadata)),
-        )
```
kespinola, line 68: This inner_join will require that each metadata has a listing_receipt which isn't the case so all the metadata that don't have a receipt will be tossed out.

kespinola (changes requested):

crates/core/src/db/queries/metadatas.rs
kespinola: What we need to be able to do is add the join conditionally with the filter so if you put that you want nfts that are listed we can do inner join for all.

PR #224, opened 2022-03-11
Left Join for Attributes Filter on NFTs

kespinola:
### Changes
- Instead of sub query left_outer join the attributes table and add filter to composite view of the data. The distinct tosses out dups and returns single copy of the nft.

Adding more filters further narrows the data set and making the query faster.

Screen Shot 2022-03-11 at 6 55 51 AM


dandlezzz (approved):
good.

PR #225, opened 2022-03-11
Inner Join for Attribute Groups

kespinola:


crates/graphql/src/schema/objects/creator.rs
```diff
@@ -30,15 +31,13 @@ impl Creator {
...
-            .filter(metadata_creators::creator_address.eq(&self.address))
-            .load(&conn)
-            .context("Failed to load metadata creators")?;
-
         let metadata_attributes: Vec<models::MetadataAttribute> = attributes::table
+            .inner_join(
+                metadata_creators::table
+                    .on(attributes::metadata_address.eq(metadata_creators::metadata_address)),
+            )
+            .filter(metadata_creators::creator_address.eq(&self.address))
```
ray-kast, line 39: does this need a `.unique()`?

kespinola: No because we want to count each occurrence.

ray-kast (approved):

PR #231, opened 2022-03-12
Index Metadata Creator Position and First Verified Creator on Attributes

kespinola:
### Changes

- Index the position of the creator in the metadata creators array. Paying out royalties for ah sales require that the position of the accounts being payed out with sales matches that of the metadata so we need to track the position of the creator in our index.
- Index the first verified creator on the attributes to eliminate doing a nest loop on the query when finding attribute groups for a creator.

### DIFF
--- a/crates/indexer/src/accountsdb/accounts/metadata.rs
+++ b/crates/indexer/src/accountsdb/accounts/metadata.rs
@@ -55,19 +55,17 @@ pub(crate) async fn process(client: &Client, key: Pubkey, meta: MetadataAccount)
         .await
         .context("Failed to dispatch metadata JSON job")?;
 
-    for (position, creator) in meta
-        .data
-        .creators
-        .iter()
-        .flatten()
-        .enumerate()
-    {
+    for (position, creator) in meta.data.creators.iter().flatten().enumerate() {
         let row = MetadataCreator {
             metadata_address: Owned(addr.clone()),
             creator_address: Owned(bs58::encode(creator.address).into_string()),
             share: creator.share.into(),
             verified: creator.verified,
-            position: Some(position.try_into().context("Position was too big to store")?),
+            position: Some(
+                position
+                    .try_into()
+                    .context("Position was too big to store")?,
+            ),
         };
 
         client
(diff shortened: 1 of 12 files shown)


crates/indexer/src/http/metadata_json.rs
```diff
@@ -419,6 +429,8 @@ pub async fn process<'a>(client: &Client, meta_key: Pubkey, uri_str: String) ->
         .chain(id.arweave.map(|a| a.0.to_vec()))
         .collect();
     let addr = bs58::encode(meta_key).into_string();
+    let first_verified_creator =
+        first_verified_creator.map(|address| bs58::encode(address).into_string());
```
ray-kast, line 433: This logic should occur in `process_attributes`

crates/indexer/src/accountsdb/accounts/metadata.rs
```diff
@@ -42,16 +47,27 @@ pub(crate) async fn process(client: &Client, key: Pubkey, meta: MetadataAccount)
         .context("Failed to insert metadata")?;
 
     client
-        .dispatch_metadata_json(key, meta.data.uri.trim_end_matches('\0').to_owned())
+        .dispatch_metadata_json(
+            key,
+            first_verified_creator,
+            meta.data.uri.trim_end_matches('\0').to_owned(),
+        )
```
ray-kast, line 54: Due to the way HTTP indexer jobs are deduplicated currently, this may be an issue and require discussion and/or refactoring.

ray-kast (changes requested):

crates/indexer/src/http/metadata_json.rs
kespinola: Don't have access to the creators in process_attributes. Need to grab at the time indexing the metadata like the address is being passed in currently.

ray-kast: Can't we just pass the un-stringified version through to process_attributes and stringify it somewhere else?  This particular function it's in mainly serves as parsing and dispatch logic for the other process functions.

ray-kast: Ended up eating my words, the fix for the other comment involved moving this logic back into this function.

kespinola (commented):
LG2M

ray-kast (approved):

Issue #227, opened 2022-03-11
Add configuration for asset CDN


ray-kast:

ray-kast:
Closed by #229, #243, #247

PR #247, opened 2022-03-15
adding dynamic cdn selection using md5, added documentation in nft im…

mpwsh:
- Added dynamic CDN selection
- Added some docs on the graphql image width parameter
- added `ASSET_PROXY_COUNT=5` to `.env` file

when `ASSET_PROXY_COUNT` is set to 1, asset endpoint url will be always `assets.holaplex.com`.
If raised, it will start adding a number to the endponit url: Ex `assets3.holaplex.com`

### DIFF
--- a/.env
+++ b/.env
@@ -2,4 +2,4 @@ ARWEAVE_URL=https://arweave.net/
 IPFS_CDN=https://ipfs.cache.holaplex.com/
 ARWEAVE_CDN=https://arweave.net/
 ASSET_PROXY_ENDPOINT=https://assets.holaplex.com/
-ASSET_PROXY_COUNT=1
+ASSET_PROXY_COUNT=5
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 ARWEAVE_URL=https://arweave.net/
 IPFS_CDN=https://ipfs.cache.holaplex.com/
 ARWEAVE_CDN=https://arweave.net/
-ASSET_PROXY_ENDPOINT=https://assets.holaplex.com/
+ASSET_PROXY_ENDPOINT=https://assets[n].holaplex.com/
 ASSET_PROXY_COUNT=5
--- a/.env
+++ b/.env
@@ -2,3 +2,4 @@ ARWEAVE_URL=https://arweave.net/
 IPFS_CDN=https://ipfs.cache.holaplex.com/
 ARWEAVE_CDN=https://arweave.net/
 ASSET_PROXY_ENDPOINT=https://assets.holaplex.com/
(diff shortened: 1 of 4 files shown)


.env
```diff
@@ -2,3 +2,4 @@ ARWEAVE_URL=https://arweave.net/
 IPFS_CDN=https://ipfs.cache.holaplex.com/
 ARWEAVE_CDN=https://arweave.net/
 ASSET_PROXY_ENDPOINT=https://assets.holaplex.com/
```
kespinola, line 4: We should put the delimiter into the asset_proxy.

`ASSET_PROXY_ENDPOINT=https://assets%d.holaplex.com/`

ray-kast: It's worth noting dynamic format strings in Rust are not possible, so it would be better to use a descriptive and unique placeholder string for search-and-replace.

ray-kast: It's worth noting dynamic format strings in Rust are not possible, so it would be better to use a descriptive and unique placeholder string for search-and-replace.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -145,27 +145,36 @@ impl Nft {
...
-                assets_cdn,
-                Base64Display::with_config(&asset.arweave.unwrap().0, base64::URL_SAFE_NO_PAD),
-                width as i32
-            )
+            let cid =
+                Base64Display::with_config(&asset.arweave.unwrap().0, base64::URL_SAFE_NO_PAD)
+                    .to_string();
+
+            let rem = md5::compute(&cid).to_vec()[0].rem_euclid(cdn_count);
+            let assets_cdn = if rem != 0 {
```
ray-kast, line 163: Clippy wants these if/else branches reversed to make the comparison a `==` instead.

ray-kast (changes requested):

.env
mpwsh: we still need to add the ASSET_PROXY_COUNT or similar to scale without the need to touch the code

ray-kast: Yeah, i think having `ASSET_PROXY_COUNT` and some kind of template string for `ASSET_PROXY_ENDPOINT` would be the ideal solution.

mpwsh: place holder added. Also the marketplaces is broken due to the double forward slash in the urls. this commit fixes that as well

ray-kast (approved):

mpwsh:
Closes #227

mpwsh:
closes [marketplace issue 105](https://github.com/holaplex/marketplace/issues/105)

PR #252, opened 2022-03-17
add dumb connection timing

dandlezzz:
dont merge me

### DIFF
--- a/crates/graphql/src/schema/query_root.rs
+++ b/crates/graphql/src/schema/query_root.rs
@@ -102,7 +102,11 @@ impl QueryRoot {
             ));
         }
 
+        let timestamp_1 = Local::now().format("%Y-%m-%dT%H:%M:%S");
+        debug!("attempting to connect to DB: {:?}", timestamp_1);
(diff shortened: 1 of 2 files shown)


crates/graphql/src/schema/query_root.rs
```diff
@@ -102,7 +102,11 @@ impl QueryRoot {
             ));
         }
 
+        let timestamp_1 = Local::now().format("%Y-%m-%dT%H:%M:%S");
+        debug!("attempting to connect to DB: {:?}", timestamp_1);
         let conn = context.db_pool.get().context("failed to connect to db")?;
```
ray-kast, line 107: This won't actually catch pool acquisitions, I can chat about how to look for that.

ray-kast (changes requested):

crates/graphql/src/schema/query_root.rs
dandlezzz: docs say:

Retrieves a connection from the pool.

Waits for at most the configured connection timeout before returning an error.


WHO IS LYING TO ME?

PR #263, opened 2022-03-21
limit creator filter to verified

dandlezzz:
This is the source of a lot of strange little bugs, notably this PR closes https://github.com/holaplex/solana-indexer/issues/262


kespinola (approved):
Do we have index on verified? Maybe a joint index creator_address, verified is quickest lookup path.

dandlezzz:
@kespinola one step ahead of ya, adding this to database; create index concurrently mcvrfd_idx on metadata_creators  (creator_address, verified);

PR #266, opened 2022-03-22
Marketplace and auction house stats.

ray-kast:


kespinola (changes requested):
Based on our slack convo floor should be the lowest listed nft.

dandlezzz (approved):
Thanks for making models!

PR #268, opened 2022-03-23
Fixups to indexer consumers.

ray-kast:
Items in this PR:
- Send a `basic.reject` for messages where a worker panic occurred
- Add a configurable timeout to the HTTP indexers
- Optimistically attempt to reuse HTTP clients

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,6 @@
 ARWEAVE_URL=https://arweave.net/
 IPFS_CDN=https://ipfs.cache.holaplex.com/
 ARWEAVE_CDN=https://arweave.net/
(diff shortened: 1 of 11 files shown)


kespinola (approved):
Can you share what led to this change? Add description to the PR for the record books.

ray-kast:
> Can you share what led to this change? Add description to the PR for the record books.

Nothing led to it per-se, but while diagnosing some other stuff I added this in to see if it helped.  This should rarely happen if ever, but essentially the missing reject call occurs if a message handler not only fails but outright panics.

PR #286, opened 2022-03-24
twitter handle indexing

imabdulbasit:


crates/core/migrations/2022-03-24-181306_twitter_handle_name_service_table/up.sql
```diff
@@ -0,0 +1,5 @@
+create table twitter_handle_name_services (
+    address                      varchar(48)    primary key,
+    wallet_address               varchar(48)    not null,
+    twitter_handle               text           not null                      
+);
```
kespinola, line 5: We'll want index on wallet_address and twitter_handle.

kespinola (commented):
so good. clippy wants some stuff.

crates/core/migrations/2022-03-24-181306_twitter_handle_name_service_table/up.sql
imabdulbasit: yeah I gotta test this first, I'll add indexes and slot thing to it

crates/indexer/src/accountsdb/accounts/name_service.rs
```diff
@@ -0,0 +1,82 @@
...
+    let values = TwitterHandle {
+        address: Owned(pubkey.clone()),
+        wallet_address: Owned(wallet.to_string()),
+        twitter_handle: Owned(th.handle),
+        slot: slot.try_into()?,
+    };
+
+    match rows.get(0) {
+        Some(indexed) => {
+            if incoming_slot > indexed.slot {
```
ray-kast, line 50:
```suggestion
        Some(indexed) if incoming_slot > indexed.slot => {
```

crates/indexer/src/accountsdb/accounts/name_service.rs
```diff
@@ -0,0 +1,82 @@
...
+                            twitter_handle_name_services::table
+                                .filter(twitter_handle_name_services::address.eq(pubkey)),
+                        )
+                        .set(&values)
+                        .execute(db)
+                    })
+                    .await
+                    .context("failed to update twitter handle")?;
+            }
+        },
```
ray-kast, line 64:
```suggestion
        },
        Some(_) => (),
```

ray-kast (changes requested):
@imabdulbasit let me know when you get the deserialization error worked out

imabdulbasit:
> @imabdulbasit let me know when you get the deserialization error worked out

Its because of incorrect hashed data so nothing much can be done

ray-kast (approved):

PR #293, opened 2022-03-25
update nfts query to get nfts with outstanding (open) offers of user

alchemistgo87:
This is an update in backend design to solve this issue: https://github.com/holaplex/marketplace/issues/88

We need to fetch nfts on which the user has outstanding (open) bids/offers.


dandlezzz:
This looks great @alchemistgo87 ! I think we are missing a few indexes for this query which we can handle in a separate PR. 

Maybe
`create index concurrently open_offers_idx on bid_receipts (buyer, purchase_receipt, canceled_at) where buyer is null and canceled_at is null;`

ray-kast:
@alchemistgo87 Needs a pass from `cargo fmt`.

ray-kast (approved):

PR #304, opened 2022-03-28
update sample config

dandlezzz:
Update sample config to include Cardinal Programs.


ray-kast:
Cherry-picking onto #303

PR #308, opened 2022-03-28
Release - Project Rename, Twitter Handle Indexing, Stats, Metadata JSON Indexing

kespinola:


ray-kast (changes requested):
Currently impossible to tell if staging is stable because our staging validator is down.  Once we can get that back up and verify the changes are working we can merge.

PR #312, opened 2022-03-29
Add remote configuration for Geyser plugins.

### LINKED ISSUE #310: Add the ability to remotely advise new Geyser plugin configurations.
Currently updating the configuration of a validator requires a restart for the Geyser plugin to reload its internal configuration.  It would be preferable if a connection could be established and maintained to allow remotely hot-reloading the plugin configuration.

ray-kast:
Closes #310.


kespinola (commented):
looking good.

.env.dev
```diff
@@ -4,3 +4,7 @@ ADDRESS=[::]:3000
 DATABASE_URL=postgres://postgres:holap1ex@localhost:5337/holaplex-indexer
 DATABASE_NAME=holaplex-indexer
 DATABASE_PASSWD=holap1ex
+
+CONFIG_DATABASE_URL=postgres://postgres:holap1exGEYS3R@localhost:5338/holaplex-indexer-config
```
kespinola, line 8: What do you think of loading the config from disk to simplify its dependencies but then manage that file using a kubernetes config map. Changing the config is updating the configmap and cycling the pods. can even put together gitops flow for it. 

We should do some interpolation at boot of the config service to switch write secret config like amp url based on environment variables of the pod.

ray-kast: The design intent here is that this is a GraphQL subscription server that we deploy once and restart as infrequently as possible.  Additionally we have multiple validators a single config server needs to manage, which would get pretty clunky to manage with a config map.

ray-kast: RE: removing postgres, I can look into ways to make this more accessible such as allowing the plugin to still load config from disk

kespinola: We don't need realtime updates I advocate for rest api that get polled by the plugin then can redeploy graph service safely without worrying about cutting off a subscription.

ray-kast: I can pivot to that once we have k8s working.  For now I'm gonna continue with the strategy I already spec'd out since most components will be reused either way.

PR #320, opened 2022-03-30
[Fix] Removing Creators From Marketplace

kespinola:
### Issues

Marketplace owners are unable to remove a creator from the marketplace. The json file is being correctly updated but the indexer was only handling insertions.

### Changes
Delete any store_creators from the index that aren't present in the latest copy of the store config.

### DIFF
--- a/crates/indexer/src/http/store_config.rs
+++ b/crates/indexer/src/http/store_config.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use indexer_core::db::{
-    insert_into,
+    delete, insert_into,
     models::{StoreConfigJson, StoreCreator},
     tables::{store_config_jsons, store_creators},
 };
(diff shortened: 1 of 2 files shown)


kespinola:
Screen Shot 2022-03-30 at 9 14 48 AM

Test of updating my local db when connected to mainnet store_config queue.

crates/indexer/src/http/store_config.rs
```diff
@@ -112,22 +112,40 @@ pub async fn process(client: &Client, config_key: Pubkey, uri_str: String) -> Re
...
+                let remove_creators = store_creators::table
+                    .filter(store_creators::store_config_address.eq(addr.clone()))
+                    .select(store_creators::creator_address)
+                    .get_results::<String>(db)
+                    .unwrap_or_else(|_| Vec::new())
+                    .into_iter()
+                    .filter(|address| !creators.clone().into_iter().any(|c| &c.address == address))
+                    .collect::<Vec<_>>();
+
+                db.build_transaction().read_write().run(|| {
```
kespinola, line 124: run the delete and upsert in a transaction so all rollbacks together in the case of an issue.

ray-kast (approved):
There are a couple other tables where we should also do this.  Metadata creators and JSON attributes are the first that come to mind.

PR #326, opened 2022-03-30
NFTs Activity Query

kespinola:
## Goal
A history of listings and purchases for an NFT.


kespinola (commented):
Nft activity needs to also return the auction house. The client should filter out activity not associated to the current marketplace based on the marketplace.address.auctionHouse

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -86,6 +89,43 @@ pub struct NftOwner {
...
+            price,
+            created_at,
+            wallets,
+            activity_type,
+        }: models::NftActivity,
+    ) -> Result<Self, Self::Error> {
+        Ok(Self {
+            address,
+            metadata,
+            auction_house,
```
kespinola, line 120: Likely want to conver this to the auction house object so holaplex.com can show the marketplace associated to the activity.

crates/core/src/db/models.rs
```diff
@@ -263,6 +263,38 @@ pub struct Nft {
     pub image: Option<String>,
 }
 
+/// Union of `listing_receipts` and `purchase_receipts` for an NFTActivity
```
ray-kast, line 266:
```suggestion
/// Union of `listing_receipts` and `purchase_receipts` for an `NFTActivity`
```

ray-kast (changes requested):

ray-kast (approved):

PR #327, opened 2022-03-31
Revert "index for token_account mint addr"

kespinola:
Reverts holaplex/indexer#319

### DIFF
--- a/crates/core/migrations/2022-03-30-160537_add_index_to_token_accounts__mint_addr/down.sql
+++ b/crates/core/migrations/2022-03-30-160537_add_index_to_token_accounts__mint_addr/down.sql
@@ -1,1 +1,0 @@
-drop index if exists token_account_mint_addr_idx;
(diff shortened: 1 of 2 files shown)


kespinola:
```
    Column     |            Type             | Collation | Nullable | Default
---------------+-----------------------------+-----------+----------+---------
 address       | character varying(48)       |           | not null |
 mint_address  | character varying(48)       |           | not null |
 owner_address | character varying(48)       |           | not null |
 amount        | bigint                      |           | not null | 0
 updated_at    | timestamp without time zone |           | not null | now()
 slot          | bigint                      |           |          |
Indexes:
    "token_accounts_pkey" PRIMARY KEY, btree (address)
    "tok_amt_idx" btree (amount)
    "tok_mint_idx" btree (mint_address)
    "tok_own_idx" btree (owner_address)
    "token_account_amt_mint_idx" btree (mint_address, amount) WHERE amount = 1
```

We already have a btree index for mint_address if we want to switch to hash then should drop the btree?

imabdulbasit (approved):

PR #335, opened 2022-04-02
Trying out indexer with Cardinal programs

jpbogle:
Just opening a draft as a WIP and had some questions i need to followup on so gonna be looking at those before merging

1. do account updates come on deletes? it seems im not handling deletes correctly - also need to understand the topics and what is on `cardinal` topic / queue_suffix since it seems like it just get everything
2. need to look more deeply into the graphql endpoints (switching to Hasura) because I think our data model is going to use Union types since the model is 1 token-manager -> n invalidators, and invalidators can be of multiple types - not 100% on the best model for this in pg
3. need to understand how migrations work, i guess for adding new fields its easy if we want to migrate the whole account on solana we could still map it to the same PG table etc - in our state we may want a concept of "only listed on `holaplex` right now they are just global
4. need to understand how the indexer works if a specific account change is failed to process, presumbly its still on rabbitMQ so it will attempt to process again - what about if accountsDB fails to put it on rabbitMQ
5. Does/ should the UI wait for the full propogation all the way to pg to before things show up or does it optimistically re-render something assuming it will go through


crates/indexer/src/geyser/programs/use_invalidator.rs
```diff
@@ -0,0 +1,21 @@
...
+    // use_invalidator::process(client, update.key, use_invalidator).await
+    let use_invalidator_discriminator: &[u8] =
+        &hash("account:UseInvalidator".as_bytes()).to_bytes()[..8];
+    let account_discriminator = &update.data[..8];
+    if account_discriminator == use_invalidator_discriminator {
+        let use_invalidator: UseInvalidator =
+            UseInvalidator::try_deserialize(&mut update.data.as_slice())
+                .context("Failed to deserialize use_invalidator")?;
+        use_invalidator::process(client, update.key, use_invalidator).await?
+    }
```
jpbogle, line 19: using account discriminator here I think this is a good approach, though a name change of the account would obviously screw this up so seems maybe hacky in that sense but also a name change is not backwards compatible so we wouldnt do that

dandlezzz:
1) No, we don't get accountUpdates on deletes. We could experiment by listening to instructions, ideally we find a better way to account for  the business-logic of a delete. Either we exclude some accounts in queries or somehow mark theme deleted. Likely have to talk this one over.

2) Sounds like a case for polymorphism, happy to think through this case with you too.

3) Let's try and treat this as a global solana cardinal index. We should be able to find a way for marketplaces to scope their queries on the entire data set.

4) We have two validators for redundancy, the likelihood of both of them missing a message is quite low, (not sure of a number here, but we haven't _really_ seen it). Should that happen, we can trigger a restart to try and get the message from historical data. If we fail at processing, we have a dead-letter and retry system that you can tap into. @ray-kast can help you with that.

5) Really up to the use case here, Accounts DB is very fast but optimistic rendering is always speedier 😄 .

crates/core/migrations/2022-04-02-083109_create_token_manager_table/up.sql
```diff
@@ -0,0 +1,64 @@
+-- Your SQL goes here
```
ray-kast, line 1:
```suggestion
```

crates/core/migrations/2022-04-02-083109_create_token_manager_table/down.sql
```diff
@@ -0,0 +1,5 @@
+-- This file should undo anything in `up.sql`
```
ray-kast, line 1:
```suggestion
```

ray-kast (changes requested):
Looks good so far

jpbogle:
notes from talk

1. do account updates come on deletes? it seems im not handling deletes correctly - also need to understand the topics and what is on cardinal topic / queue_suffix since it seems like it just get everything
- Check in graphql query that expiration || duration + state_changed_at  > Date.now()
- index expiration, duration

2. need to look more deeply into the graphql endpoints (switching to Hasura) because I think our data model is going to use Union types since the model is 1 token-manager -> n invalidators, and invalidators can be of multiple types - not 100% on the best model for this in pg
- polymophic table which will hold both kinds of invalidators

3. need to understand how migrations work, i guess for adding new fields its easy if we want to migrate the whole account on solana we could still map it to the same PG table etc - in our state we may want a concept of "only listed on holaplex right now they are just global
- same as account migration

4. need to understand how the indexer works if a specific account change is failed to process, presumbly its still on rabbitMQ so it will attempt to process again - what about if accountsDB fails to put it on rabbitMQ
- redudancy and shouldnt happen much if at all (answered above)

5. Does / should the UI wait for the full propogation all the way to pg to before things show up or does it optimistically re-render something assuming it will go through
- answered above

diesel.toml
```diff
@@ -0,0 +1,5 @@
+# For documentation on how to configure this file,
```
kespinola, line 1: looks like migration and diesel file added to the root from some init command of desiel. 

We have `./diesel.sh` to wrap the cli and help working with the workspaces.

ray-kast:
@jpbogle just fyi dev appears to have some conflicting changes now

jpbogle:
@ray-kast fixed merge conflicts and opened draft for review --

@dandlezzz we are actually able to keep ihstoricals in our case because we have timestamp on-chain so if we receive those updates again, it will just ovverwrite the existing one

crates/indexer/src/geyser/accounts/cardinal_token_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+                    cardinal_use_invalidators::use_invalidator_collector.nullable(),
+                    cardinal_use_invalidators::use_invalidator_usages.nullable(),
+                    cardinal_use_invalidators::use_invalidator_use_authority.nullable(),
+                    cardinal_use_invalidators::use_invalidator_total_usages.nullable(),
+                    cardinal_use_invalidators::use_invalidator_extension_payment_amount.nullable(),
+                    cardinal_use_invalidators::use_invalidator_extension_payment_mint.nullable(),
+                    cardinal_use_invalidators::use_invalidator_extension_usages.nullable(),
+                    cardinal_use_invalidators::use_invalidator_max_usages.nullable(),
+                ))
+                .load(db)
```
kespinola, line 108: nb: Since you are looking for 1 result you can do `find(1).first(db)` this will return an Err() if the record doesn't exist can check that instead of the len.

https://github.com/diesel-rs/diesel/blob/master/diesel_tests/tests/find.rs#L112

kespinola (approved):
Have we chatted about the number of accounts that are about to be indexed and the rate of change? Lets chat in are shared channel just want to get a sense of the load.

crates/core/migrations/2022-04-02-083109_create_cardinal_token_manager_tables/up.sql
```diff
@@ -0,0 +1,125 @@
...
+);
+
+create index if not exists cardinal_token_managers_issuer_idx
+on cardinal_token_managers (issuer);
+
+create table cardinal_token_manager_invalidators (
+    token_manager_address   varchar(48) not null,
+    invalidator             varchar(48) not null,
+    primary key (token_manager_address, invalidator),
+    foreign key (token_manager_address) references cardinal_token_managers (address)
```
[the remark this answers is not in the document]
jpbogle: we are getting these from the list of invalidators on the token_manager account itself, so we actually insert the row to token_managers table first and then add these

crates/indexer/src/geyser/accounts/cardinal_paid_claim_approver.rs
```diff
@@ -0,0 +1,45 @@
...
+        ),
+        paid_claim_approver_collector: Owned(
+            bs58::encode(paid_claim_approver.collector).into_string(),
+        ),
+        paid_claim_approver_payment_amount: paid_claim_approver.payment_amount.try_into()?,
+        paid_claim_approver_payment_mint: Owned(
+            bs58::encode(paid_claim_approver.payment_mint).into_string(),
+        ),
+    };
+    debug!("Paid Claim Approver {:?}", row);
```
ray-kast, line 30:
```suggestion
    trace!("Paid Claim Approver {:?}", row);
```

crates/indexer/src/geyser/accounts/cardinal_time_invalidator.rs
```diff
@@ -0,0 +1,63 @@
...
+            .transpose()?,
+        time_invalidator_max_expiration: time_invalidator
+            .max_expiration
+            .map(|e| NaiveDateTime::from_timestamp(e, 0)),
+        time_invalidator_disable_partial_extension: time_invalidator
+            .disable_partial_extension
+            .map(TryFrom::try_from)
+            .transpose()?,
+    };
+    debug!("Time invalidator {:?}", row);
```
ray-kast, line 48:
```suggestion
    trace!("Time invalidator {:?}", row);
```

crates/indexer/src/geyser/accounts/cardinal_token_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+                .map(|k| Owned(bs58::encode(k).into_string())),
+
+            ////////////// claim approver //////////////
+            paid_claim_approver_payment_amount: current_token_manager
+                .paid_claim_approver_payment_amount
+                .try_into()?,
+            paid_claim_approver_payment_mint: current_token_manager
+                .paid_claim_approver_payment_mint
+                .clone()
+                .map(|k| Owned(k)),
```
ray-kast, line 148: Generally, `.map(|x| Owned(x))` can be replaced with `.map(Owned)`

crates/indexer/src/geyser/accounts/cardinal_token_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+            .receipt_mint
+            .map(|k| Owned(bs58::encode(k).into_string())),
+        claim_approver: token_manager
+            .claim_approver
+            .map(|k| Owned(bs58::encode(k).into_string())),
+        transfer_authority: token_manager
+            .transfer_authority
+            .map(|k| Owned(bs58::encode(k).into_string())),
+    };
+    debug!("Processing token manager {:?}", row);
```
ray-kast, line 266:
```suggestion
    trace!("Processing token manager {:?}", row);
```

crates/indexer/src/geyser/accounts/cardinal_token_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+                .set(&row)
+                .execute(db)
+        })
+        .await
+        .context("Failed to insert TokenManager")?;
+
+    let mut invalidator_strings = Vec::new();
+    for invalidator in token_manager.invalidators {
+        invalidator_strings.push(bs58::encode(invalidator).into_string())
+    }
```
ray-kast, line 284:
```suggestion
    let invalidator_strings: Vec<_> = token_manager.invalidators.into_iter().map(|i| bs58::encode(i).into_string()).collect();
```

ray-kast (changes requested):
Overall looking pretty good!

kespinola (approved):

ray-kast (approved):

PR #343, opened 2022-04-06
Show listed nfts on top for 'ALL' filter

alchemistgo87:
With reference to https://github.com/holaplex/marketplace/issues/145, this PR changes the order of nfts in the 'All' filter by showing listed nfts first in the list.


crates/core/src/db/queries/metadatas.rs
```diff
@@ -89,7 +92,9 @@ pub fn list(
                 metadata_jsons::image,
             ))
             .distinct()
-            .order(metadatas::address.asc())
+            .order((listing_receipts::price.desc(), metadatas::name.asc()))
```
ray-kast, line 95: How does this perform at scale?  We've had problems with this particular line before.

kespinola:
```
Indexes:
    "listing_receipts_pkey" PRIMARY KEY, btree (address)
    "listing_receipts_address_index" btree (address)
    "listing_receipts_auction_house_index" btree (auction_house)
    "listing_receipts_metadata_index" btree (metadata)
    "listing_receipts_purchase_receipt_idx" btree (purchase_receipt)
```

```
Indexes:
    "metadatas_pkey" PRIMARY KEY, btree (address)
    "meta_name" btree (name)
    "metadatas_mint_address_index" hash (mint_address)
```

We have index on metadata name. Definitely need one on lr price. can you add a migration? just reminder when adding migrations make sure to have db url set to your local db.

still need to do performance comparison but index needed to give it a shot.

ray-kast (changes requested):
Add `#[allow(clippy::too_many_lines)]` between lines 53 and 54 of `queries/metadatas.rs` to silence clippy.  Normally I'd request the function be broken up, but since this is assembling a monolithic Diesel query that's likely not gonna be possible without inducing tech debt.

crates/core/src/db/queries/metadatas.rs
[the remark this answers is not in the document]
alchemistgo87: @kespinola The issue is related to "distinct" which demands the order by field to be one of the distinct fields - https://stackoverflow.com/questions/9795660/postgresql-distinct-on-with-different-order-by/9796104#9796104

kespinola (approved):
Pulled branch and patched it up. Ready to go.

ray-kast (approved):

PR #344, opened 2022-04-06
Query Graph Connection Enriched with Twitter Profile

kespinola:
### Goals

Clients can query graph connections and have access to twitter info of the wallet when linked with bonfida.

### Changes
- Add root query for `connections(from, to, limit, offset)`
- wallet and creator root queries include profile when linked

### DIFF
--- a/crates/core/src/db/queries/graph_connection.rs
+++ b/crates/core/src/db/queries/graph_connection.rs
@@ -1,0 +1,48 @@
+//! Query utilities for `graph_connections` table.
+
+use diesel::{
+    pg::Pg,
+    serialize::ToSql,
+    sql_query,
+    sql_types::{Array, Int4, Text},
+};
+
+use crate::{
+    db::{models::TwitterEnrichedGraphConnection, Connection},
+    error::Result,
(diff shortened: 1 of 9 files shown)


crates/core/src/db/queries/graph_connection.rs
```diff
@@ -0,0 +1,48 @@
...
+    FROM graph_connections gc
+    LEFT JOIN twitter_handle_name_services fth ON gc.from_account = fth.wallet_address
+    LEFT JOIN twitter_handle_name_services tth ON gc.to_account = tth.wallet_address
+    WHERE ($1 = '{}' OR from_account = ANY($1)) AND ($2 = '{}' OR to_account = ANY($2))
+    ORDER BY connection_address
+    LIMIT $3 OFFSET $4;
+ -- $1: from::text[]
+ -- $2: to::text[]
+ -- $3: limit::integer
+ -- $4: limit::integer
```
ray-kast, line 27:
```suggestion
 -- $4: offset::integer
```

ray-kast (changes requested):

PR #345, opened 2022-04-06
NFT Charts Queries

kespinola:
## Changes
Query collection chart data from API.


kespinola (commented):
@alchemistgo87 this is looking good. you do some testing on it? query ok?

alchemistgo87:
@kespinola  The query is throwing an error for both floor_price and average_price - "out of range integral type conversion attempted"

kespinola:
```
 GroupAggregate  (cost=15.11..18.71 rows=200 width=16) (actual time=0.170..0.194 rows=34 loops=1)
   Group Key: series.series
   ->  Merge Left Join  (cost=15.11..16.81 rows=1000 width=16) (actual time=0.164..0.175 rows=38 loops=1)
         Merge Cond: (series.series = (date_trunc('day'::text, pr.created_at)))
         ->  Sort  (cost=12.97..13.47 rows=1000 width=8) (actual time=0.050..0.052 rows=34 loops=1)
               Sort Key: series.series
               Sort Method: quicksort  Memory: 26kB
               ->  Function Scan on generate_series series  (cost=0.00..3.00 rows=1000 width=8) (actual time=0.022..0.024 rows=34 loops=1)
         ->  Sort  (cost=2.14..2.15 rows=9 width=16) (actual time=0.112..0.112 rows=8 loops=1)
               Sort Key: (date_trunc('day'::text, pr.created_at))
               Sort Method: quicksort  Memory: 25kB
               ->  Seq Scan on purchase_receipts pr  (cost=0.00..2.11 rows=9 width=16) (actual time=0.044..0.091 rows=8 loops=1)
                     Filter: ((created_at >= '2022-03-01 15:17:13'::timestamp without time zone) AND (created_at <= '2022-04-03 15:17:13'::timestamp without time zone) AND ((auction_house)::text = 'EsrVUnwaqmsq8aDyZ3xLf8f5RmpcHL6ym5uTzwCRLqbE'::text))
                     Rows Removed by Filter: 18
 Planning Time: 0.851 ms
 Execution Time: 0.293 **ms**
```

Need to look at index for the where clause of purchase_receipts and listing_receipts

ray-kast (approved):

crates/core/src/db/models.rs
```diff
@@ -1293,3 +1305,43 @@ pub struct InsBufferBundleInsKey<'a> {
     /// True if the `pubkey` can be loaded as a read-write account.
     pub is_writable: bool,
 }
+
+/// Union of `listing_receipts` and `purchase_receipts` for an auction house `Activity`
+#[derive(Debug, Clone, Queryable, QueryableByName)]
```
kespinola, line 1310: We don't need another model for Activity this can be the existing NftActivity. What we do need to do is add field to NFT activity for the nft. This should be dataloader based on metadata address and respond with nft objects so caller gets access to all the info they may want from the nft not lock into just the image, description.

```
activities(auctionHouses: []. limit: 10, offset: 0) {
  nft {
    image
    creators {
      ...
    }
    ...
  }
  price
  activity_type
}
```

This will be 2 queries but right now select then join is same number of select statements but 1 query submitted to the api. Like to try this because it uses existing objects which leads to consistent api.

crates/core/src/db/queries/activities.rs
```diff
@@ -0,0 +1,44 @@
...
+    LEFT JOIN metadatas ON act.metadata = metadatas.address
+    LEFT JOIN metadata_jsons ON act.metadata = metadata_jsons.metadata_address
+    ORDER BY act.created_at DESC;
+ -- $1: auction_houses::text[]";
+
+/// Load all activities for desired auction house address
+///
+/// # Errors
+/// This function fails if the underlying SQL query returns an error
+pub fn activities(
```
kespinola, line 36:
```suggestion
pub fn list(
```

shoud accept limit and offset as well.

crates/graphql/src/schema/objects/activity.rs
```diff
@@ -0,0 +1,46 @@
...
+            metadata_image,
+            metadata_name,
+            auction_house,
+            price: price.try_into()?,
+            created_at: DateTime::from_utc(created_at, Utc),
+            wallets,
+            activity_type,
+        })
+    }
+}
```
kespinola, line 46: Can drop when switch to current NftActivity object.

kespinola (commented):
@alchemistgo87 left notes to wrap up activities. like to use existing NftActivity object but expose nft field from it to query the preview information.

crates/graphql/src/schema/dataloaders/nft.rs
```diff
@@ -168,3 +168,13 @@ impl TryBatchFn<PublicKey<Nft>, Vec<NftActivity>> for Batcher {
...
+
+#[async_trait]
+impl TryBatchFn<PublicKey<NftActivity>, Nft> for Batcher {
+    async fn load(
+        &mut self,
+        addresses: &[PublicKey<NftActivity>],
+    ) -> TryBatchMap<PublicKey<NftActivity>, Nft> {
+        let conn = self.db()?;
+    }
+}
```
kespinola, line 180: I needed this very loader type for feed_events. Can you take this one and add here.

https://github.com/holaplex/indexer/pull/368/files#diff-beaf2efe58be82e65ef92d01fba2f303960ebe4530074a71288cbf4eb78d34e3R173

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -206,6 +206,16 @@ impl TryFrom<models::NftActivity> for NftActivity {
     }
 }
 
+#[graphql_object(Context = AppContext)]
+impl NftActivity {
+    pub async fn nft(&self, ctx: &AppContext) -> FieldResult<Nft> {
+        ctx.nft_preview_loader
+            .load(self.address.clone().into())
+            .await
+            .map_err(Into::into)
```
kespinola, line 215: nft_loader based on the example I shared for upcoming feed events.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -206,6 +206,16 @@ impl TryFrom<models::NftActivity> for NftActivity {
     }
 }
 
+#[graphql_object(Context = AppContext)]
```
kespinola, line 209: Need to drop `GraphQLObject` on the NftActivity struct and add getter methods for all the fields on the struct.

kespinola (commented):
left notes on doing the nft data loader.

crates/graphql/src/schema/query_root.rs
```diff
@@ -16,7 +17,7 @@ use tables::{
     store_config_jsons, storefronts,
 };
 
-use super::prelude::*;
+use super::{objects::nft::NftActivity, prelude::*};
```
kespinola, line 20: can pull nft:NftActivity from the existing object export. leave this as just the prelude.

kespinola (commented):
Looks great just need rebase in dev and maybe fix up the import i noted.

kespinola (commented):
LG2M 👍

PR #349, opened 2022-04-07
remove duplicate attributes

imabdulbasit:
Attributes table has no unique constraint on (metadata_address, value, trait_type) because these fields are optional so we can not use primary key constraint on them.

So, a temp table is created which contains only distinct rows from attributes table and has a UNIQUE constraint on (metadata_address, value, trait_type) . The attributes table is renamed to temp_attributes and the temp table is then renamed to attributes


imabdulbasit:
@kespinola we can rename the attributes table to temp_attributes instead of dropping so if there is any problem we can run a migration to rename it back to attributes

PR #352, opened 2022-04-07
Added connection counts for profile brief (wallet)

kevinrodriguez-io:


crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -20,16 +21,15 @@ impl Wallet {
...
     }
 
     pub fn bids(&self, ctx: &AppContext) -> FieldResult<Vec<Bid>> {
         let db_conn = ctx.db_pool.get()?;
-
         let rows: Vec<models::Bid> = bids::table
-            .select(bids::all_columns)
-            .filter(bids::bidder_address.eq(&self.address))
+            .select(bids::all_columns)      // How to avoid cloning?
+            .filter(bids::bidder_address.eq::<String>(self.address.clone().into()))
```
ray-kast, line 32: PublicKey should be usable here without an Into

ray-kast (changes requested):
PublicKey should be usable directly in SQL queries.  If not I can fix that.

crates/graphql/src/schema/objects/wallet.rs
kevinrodriguez-io: I get this when I remove the eq::<String> (just eq) and using the PublicKey directly
```
the trait bound `PublicKey<Wallet>: diesel::expression::Expression` is not satisfied
required because of the requirements on the impl of `diesel::expression::AsExpression<Text>` for `PublicKey<Wallet>`
```

ray-kast: I'll look into this.

ray-kast (approved):

crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -41,17 +41,51 @@ impl Wallet {
...
+        let db_conn = ctx.db_pool.get()?;
+
+        let count: i64 = graph_connections::table
+            .filter(graph_connections::to_account.eq(&self.address))
+            .count()
+            .get_result(&db_conn)
+            .context("Failed to count to_connections")?;
+
+        Ok(count.try_into()?)
+    }
```
kespinola, line 90: We dont have a list endpoint for wallets but if we add this needs to move to a data loader.

crates/graphql/src/schema/scalars/public_key.rs
```diff
@@ -117,6 +117,28 @@ where
...
+where
+    &'a String: indexer_core::db::expression::AsExpression<U>,
+{
+    type Expression = <&'a String as indexer_core::db::expression::AsExpression<U>>::Expression;
+
+    fn as_expression(self) -> Self::Expression {
+        (&self.0).as_expression()
+    }
+}
+
```
kespinola, line 141: sweet.

kespinola (approved):

PR #353, opened 2022-04-07
create new attributes table with unique constraint

imabdulbasit:


crates/core/src/db/schema.rs
```diff
@@ -8,7 +8,6 @@ table! {
         value -> Nullable<Text>,
         trait_type -> Nullable<Text>,
         id -> Uuid,
-        first_verified_creator -> Nullable<Varchar>,
```
ray-kast, line 11: Did we mean to remove this?

kespinola, line 11: Losing first_verified_creator? don't think we want to do that.

imabdulbasit: I forgot to add it as it was a separate migration

ray-kast (approved):
We will have to keep an eye on dev to revert the migration being edited here.  @kespinola @dandlezzz

PR #354, opened 2022-04-08
Add reverse lookup for wallet address from twitter handle

kevinrodriguez-io:
This feature allows getting a reverse lookup for handle -> wallet

### DIFF
--- a/crates/graphql/src/schema/objects/profile.rs
+++ b/crates/graphql/src/schema/objects/profile.rs
@@ -1,9 +1,11 @@
 use serde::Deserialize;
+use tables::twitter_handle_name_services;
 
 use super::prelude::*;
 
 #[derive(Debug, Clone, GraphQLObject)]
 pub struct TwitterProfile {
+    pub wallet_address: Option<String>,
     pub handle: String,
     pub profile_image_url: String,
     pub banner_image_url: String,
@@ -12,6 +14,7 @@ pub struct TwitterProfile {
 
 #[derive(Debug, Clone)]
 pub struct Profile {
+    pub wallet_address: Option<String>,
     pub handle: String,
     pub profile_image_url_lowres: String,
     pub profile_image_url_highres: String,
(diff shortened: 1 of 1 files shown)


crates/graphql/src/schema/objects/profile.rs
```diff
@@ -63,6 +67,21 @@ pub struct TwitterUserProfileResponse {
...
+            .filter(twitter_handle_name_services::twitter_handle.eq(&self.handle))
+            .load(&db_conn)
+            .context("Failed to load wallet address")?;
+        if result.len() == 0 {
+            return Ok(None);
+        }
+        let matching_item = result.get(0).unwrap();
+        let wallet_address = &matching_item.wallet_address;
+        Ok(Some(wallet_address.to_string()))
+    }
```
ray-kast, line 84: I'm confused why wallet_address is stored on the Profile struct but also a computed field.

ray-kast (changes requested):

crates/graphql/src/schema/objects/profile.rs
kevinrodriguez-io: You're right! Removed

kevinrodriguez-io: You're right! Removed

ray-kast (approved):
Looks good, although some CI steps are still failing.

crates/graphql/src/schema/objects/profile.rs
```diff
@@ -63,6 +64,21 @@ pub struct TwitterUserProfileResponse {
...
+            .filter(twitter_handle_name_services::twitter_handle.eq(&self.handle))
+            .load(&db_conn)
+            .context("Failed to load wallet address")?;
+        if result.is_empty() {
+            return Ok(None);
+        }
+        let matching_item = result.get(0).unwrap();
+        let wallet_address = &matching_item.wallet_address;
+        Ok(Some(wallet_address.to_string()))
+    }
```
kespinola, line 81: This needs to be a data loader or we have n + 1.

kevinrodriguez-io: @kespinola I'll create a ticket for this.

PR #361, opened 2022-04-11
Add wait untill complete blocks

echohtp:


.github/workflows/dispatch_to_docs.yml
```diff
@@ -8,9 +8,23 @@ jobs:
   dispatch:
     runs-on: ubuntu-latest
     steps:
-      - name: Dispatch event to docs repo
-        uses: peter-evans/repository-dispatch@v2
-        with:
+     - name: Wait for dev deploy to succeed
+       uses: lewagon/wait-on-check-action@v1.0.0
+       with:
+          ref: master
```
ray-kast, line 14: If this keys against master, you'll likely need to make the dev one a separate action, no?

(also we don't quite have k8s working yet lol)

ray-kast (dismissed):

.github/workflows/dispatch_to_docs.yml
echohtp: yes, thanks for catching the `master` / `dev`. 

i think as long as the dispatch event waits for the k8s build to finish, i'll be happy.

ray-kast (approved):

PR #362, opened 2022-04-12
Various refactors accumulated from ryans/remote-config

ray-kast:
- Updated the workspace lockfile
- Fixup to start-developing.sh
- Fixed a bug with Sender.
- Collect host and build data for identifying plugin
- Added feature flags to the core crate.
- Resolve dead-code warnings in some RabbitMQ builds
- Added customizable consumer tags.
- Improve indexer concurrency options.
- Bump Geyser interface version


crates/geyser-rabbitmq/src/plugin.rs
```diff
@@ -78,8 +78,42 @@ impl GeyserPlugin for GeyserPluginRabbitMq {
...
+                }
+
+                version = s;
+            }
+
+            // let rustc_ver = env!("META_RUSTC_VERSION");
+            // let build_host = env!("META_BUILD_HOST");
+            // let target = env!("META_BUILD_TARGET");
+            // let profile = env!("META_BUILD_PROFILE");
+            // let features = env!("META_BUILD_FEATURES");
```
kespinola, line 108: You got these. Want to delete the comments?

crates/indexer/src/lib.rs
```diff
@@ -101,10 +101,24 @@ mod runtime {
...
+                .with_connection_name(
+                    format!(
+                        "{}@{}",
+                        name,
+                        hostname::get()
+                            .context("Failed to get system hostname")?
+                            .into_string()
+                            .map_err(|_| anyhow!("Failed to parse system hostname"))?,
+                    )
+                    .into(),
```
kespinola, line 120: nice.

kespinola (approved):

crates/geyser-rabbitmq/src/plugin.rs
ray-kast: They're produced in the build.rs and passed at build-time to these variables, but I don't use 'em at the moment.  Plan is to send more detailed telemetry in the future.

PR #368, opened 2022-04-13
Create and Query Feed Events

kespinola:
### Goals

Create events for notable changes in the ecosystem associated to relevant wallets as they are processed by the indexer. Let clients query events for a wallet that are associated to the current wallet or those it is following.

## Supported Events

- [X] Nft Minted
- [x] Offer Made
- [x] Nft Listed
- [x] Nft Sold
- [x] Wallet Followed

Screen Shot 2022-04-13 at 8 43 21 AM


kespinola:
Example query for getting all event types sorted by created_at.

```graphql
  feedEvents(wallet: "CqC5dNspaUVbUz6QDbNcodpTfgbw1nZ4ra2nectDv1ak", limit: 1000, offset:0) {
    ... on MintEvent {
      createdAt
      nft {
        name
        image
	description
      }
    }
    ... on FollowEvent {
      createdAt
      connection {
	to {
           address
           profile {
             handle
           }
        }
      }
    }
    ... on PurchaseEvent {
      createdAt
      purchase {
        buyer
        seller
        price
        nft {
          name
          image
          description
        }
      }
    }
    ... on ListingEvent {
      createdAt
      listing {
        seller
        price
        nft {
          name
          image
          description
        }
      }
    }
    ... on OfferEvent {
      createdAt
      offer {
        buyer
        price
        nft {
          name
          image
          description
        }
      }
    }
  }
```

crates/indexer/src/geyser/accounts/graph.rs
```diff
@@ -22,7 +28,51 @@ pub(crate) async fn process(client: &Client, key: Pubkey, account_data: Connecti
...
+            db.build_transaction().read_write().run(|| {
+                let follow_event_exists = select(exists(
+                    follow_events::table
+                        .filter(follow_events::graph_connection_address.eq(row.address.clone())),
+                ))
+                .get_result::<bool>(db);
+
+                if Ok(true) == follow_event_exists {
+                    return Result::<_>::Ok(());
+                }
```
kespinola, line 42: Since there is a unique constraint on each event type table we could drop the select and let the transaction rollback that is already occurring when 2 consumer jobs processing the same account change for the first time.

crates/indexer/src/geyser/accounts/metadata.rs
```diff
@@ -2,16 +2,28 @@ use indexer_core::{
...
 use mpl_token_metadata::state::{Collection, Metadata as MetadataAccount, TokenStandard};
 
 use super::Client;
 use crate::prelude::*;
 
+#[derive(Clone, Copy)]
+enum FeedEventId {
+    Existing,
+    Inserted(Uuid),
+}
```
ray-kast, line 25: This (and the previous `Either` version) is isomorphic to `Option<Uuid>`.  Is there any reason we don't just do that?

ray-kast (changes requested):

crates/indexer/src/geyser/accounts/metadata.rs
kespinola: word. switched to Option.

kespinola:
Duplicate constraints still showing warning. took a stab at handling the rollback error but no luck. Can update in the future to silent the warn logs.

kristianeboe (approved):

PR #378, opened 2022-04-19
Make the validators talk to us

ray-kast:


kespinola (commented):
Can you add some notes on the structure of this? From what I can see its message sending and reading using rabbit?

PR #381, opened 2022-04-19
Add Strata token bonding reserve/supply changes

ChewingGlass:
This pull request adds a table `bonding_changes` that observes fields tracking the reserves and supply on a bonding curve.

This creates a time series table of observed reserves and supply, which we can then run a lag query on to get _changes_ in reserve/supply. These changes can give us an estimate on the price, which we can then plot.

This is a temporary solution that will be replaced as soon as transactions are flowing in, which will give us a more accurate look at pricing at each point in time.


ray-kast (changes requested):
[patch with recommended changes](https://github.com/holaplex/indexer/files/8536018/0001-Fixup-to-token-bonding-PR.patch.zip)

This is a mix of stylistic tweaks to the PR as well as resolving some minor existing tech debt issues seemingly encountered by the PR.

Other than that, looking very solid!

ray-kast (approved):

PR #385, opened 2022-04-20
Add Dialect API Post call for integration

echohtp:


crates/indexer/src/geyser/client.rs
```diff
@@ -92,4 +95,34 @@ impl Client {
...
+            .new()
+            .post("")
+            .body(Message {
+                msgType: MessageType::NftOffer,
+                data: MessageData {
+                    address: "1111111111111111111111".to_string(),
+                },
+            })
+            .send()
+            .await?;
```
kespinola, line 126: http is misleading for this client. its actually a rabbitmq client.

you aren't looking to add another producer. its something more like the db but preconfigured http_sender.

echohtp: hm, okay :+1:

crates/indexer/src/geyser/accounts/receipt.rs
```diff
@@ -120,15 +121,23 @@ pub(crate) async fn process_bid_receipt(
...
+            ))
+            .get_result::<bool>(db);
             insert_into(bid_receipts::table)
                 .values(&row)
                 .on_conflict(bid_receipts::address)
                 .do_update()
                 .set(&row)
-                .execute(db)
+                .execute(db)?;
+            if Ok(true) == bid_receipt_exists || row.purchase_receipt.is_some() {
```
ray-kast, line 134: This if statement does nothing, `return` in Rust closures is more like `next` in Ruby closures.

ray-kast (changes requested):
What are the timings on the Dialect POST requests?  If it's too slow we may need a separate HTTP indexer for that.

crates/indexer/src/geyser/accounts/receipt.rs
echohtp: whats the suggestion here @ray-kast ?

ray-kast (changes requested):
Something has gone horribly wrong here.

ray-kast:
Gonna reopen this with a less broken Git history.

PR #386, opened 2022-04-21
MeiliSearch: Metadatas and NameService index

imabdulbasit:


crates/rabbitmq/src/search_indexer.rs
```diff
@@ -12,14 +14,15 @@ pub struct Document {
...
+    #[serde(flatten)]
+    pub body: HashMap<String, Value>,
 }
 
 /// Message data for a search index job
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum Message {
     /// Upsert a document to the index
-    Upsert(Document),
+    Upsert(String, Document),
```
ray-kast, line 25:
```suggestion
    Upsert { index: String, document: Document },
```

ray-kast (changes requested):

ray-kast (approved):

PR #387, opened 2022-04-21
Log long queries 

dandlezzz:
log long queries from graphql

### DIFF
--- a/crates/graphql/src/main.rs
+++ b/crates/graphql/src/main.rs
@@ -13,7 +13,15 @@ use std::sync::Arc;
 use actix_cors::Cors;
 use actix_web::{http, middleware, web, App, Error, HttpResponse, HttpServer};
 use indexer_core::{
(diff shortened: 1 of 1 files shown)


mpwsh (approved):

crates/graphql/src/main.rs
```diff
@@ -73,7 +81,19 @@ async fn graphql(
...
+    let end = Local::now();
+    let duration = end - start;
+    let formatted_duration = duration_hhmmssfff(duration);
+
+    if duration > Duration::milliseconds(5000) {
+        warn!(
+            "long graphql request query={:?}, duration={:?}",
+            req, formatted_duration
+        );
+    }
```
kespinola, line 96: Don't want to log every response time?

```
graphq_request query={:?} duration={:?}
```

kespinola: Be cool to base64 encode the query string as well (ideally without variables) so can compare query times between requests and over time.

kespinola (approved):

PR #408, opened 2022-04-26
Fix[Feed Events]: Use constraint instead of foreign key for metadatas

kespinola:
### Changes
- Use unique constraint instead of foreign key to enforce uniqueness.


kespinola (commented):
Going to add constraints for the other event tables.

ray-kast (approved):

PR #415, opened 2022-04-28
Expose deserializer error information in the NFT.

ray-kast:


mpwsh (approved):

kespinola (approved):
Looks good. Only item is the use of the word "model" for this. Thoughts on "parser"?

ray-kast:
> Looks good. Only item is the use of the word "model" for this. Thoughts on "parser"?

Yeah, can do that.  The `model` column has been in the JSON table since day 1 so I can simply rename the exposed GraphQL field.

PR #421, opened 2022-04-29
Query NFTs by collection

imabdulbasit:
- Collection root query which takes collection nft address as parameter
- collection data loader added in Nft

### DIFF
--- a/crates/graphql/src/schema/objects/collection.rs
+++ b/crates/graphql/src/schema/objects/collection.rs
@@ -1,0 +1,58 @@
+use super::prelude::*;
+
+#[derive(Debug, Clone)]
+pub struct Collection {
+    pub address: String,
+    pub name: String,
+    pub mint_address: String,
+    pub description: Option<String>,
+    pub image: Option<String>,
+}
+
+#[graphql_object(Context = AppContext)]
+impl Collection {
+    pub fn address(&self) -> &str {
+        &self.address
+    }
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn mint_address(&self) -> &str {
+        &self.mint_address
+    }
+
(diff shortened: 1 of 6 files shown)


ray-kast (approved):

crates/graphql/src/schema/objects/collection.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub fn image(&self) -> Option<&str> {
+        self.image.as_deref()
+    }
+
+    pub async fn nfts(&self, ctx: &AppContext) -> FieldResult<Vec<CollectionNft>> {
+        ctx.collection_nfts_loader
+            .load(self.address.clone().into())
+            .await
+            .map_err(Into::into)
+    }
```
kespinola, line 64: Need to get limit and offset as arguments for this field or going to be returning thousands of nfts.

kespinola (changes requested):
Like to release with limits and offset for the nfts field on collections.

kespinola:
Chatted in standup Thursday but for v1 of filtering nfts by collections about allowing subquery but returning all the addresses of nfts belonging to a collection to filter the existing query is going result of 10K plus address being returned for a pfp project for example. Does this sound right @imabdulbasit?

imabdulbasit:
> Chatted in standup Thursday but for v1 of filtering nfts by collections about allowing subquery but returning all the addresses of nfts belonging to a collection to filter the existing query is going result of 10K plus address being returned for a pfp project for example. Does this sound right @imabdulbasit?

We already pass collection mint address as filter. So if the filter passes collection metadata address, we need the mint address and for that we need a separate query which will return one mint address of a collection and use the same filter like we have.

ray-kast (changes requested):

PR #427, opened 2022-05-02
Add creators field in charts object for analytics

alchemistgo87:
Addresses - https://github.com/holaplex/marketplace/issues/208

- Add creators argument in the charts query
- Update charts query
- Update activities query with detailed wallet information including twitter profile data.
- Add index for buyer, seller columns in purchase_receipts table

### DIFF
--- a/crates/graphql/src/schema/objects/chart.rs
+++ b/crates/graphql/src/schema/objects/chart.rs
@@ -7,7 +7,7 @@ use super::prelude::*;
 #[derive(Debug, Clone)]
 pub struct PriceChart {
     pub auction_houses: Vec<PublicKey<AuctionHouse>>,
-    pub creators: Vec<PublicKey<Creator>>,
+    pub creators: Option<Vec<PublicKey<Creator>>>,
     pub start_date: DateTime<Utc>,
     pub end_date: DateTime<Utc>,
 }
@@ -53,8 +53,8 @@ impl PriceChart {
         let conn = ctx.shared.db.get()?;
         let rows = charts::average_prices(
             &conn,
-            &self.auction_houses,
             &self.creators,
+            &self.auction_houses,
             self.start_date.naive_utc(),
             self.end_date.naive_utc(),
         )?;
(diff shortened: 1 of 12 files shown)


kespinola (changes requested):
looks good. I'd like to drop the charts field off of creators for now because it will lead to an n+1 if we request the charts from a list of creators.

kespinola (approved):

kespinola (changes requested):
Going to look into making creators an optional field.

kespinola (approved):

crates/graphql/src/schema/objects/chart.rs
```diff
@@ -1,12 +1,13 @@
...
-use objects::auction_house::AuctionHouse;
+use objects::{auction_house::AuctionHouse, creator::Creator};
 use scalars::{PublicKey, U64};
 
 use super::prelude::*;
 
 #[derive(Debug, Clone)]
 pub struct PriceChart {
     pub auction_houses: Vec<PublicKey<AuctionHouse>>,
+    pub creators: Option<Vec<PublicKey<Creator>>>,
```
ray-kast, line 10: Any reason this isn't just a `Vec<Creator>`?

ray-kast (changes requested):

crates/graphql/src/schema/objects/chart.rs
alchemistgo87: @ray-kast Creators field can be null in case we want to fetch charts data for the whole marketplace. And in case of a particular creater's chart data creator field needs to be populated. So, kept it optional.

kespinola: Yep. If no creators provided returned stats for the whole auction_house or houses.

ray-kast: Ah, it's a query parameter.  Got it.

ray-kast (approved):

kespinola (approved):
This is good but need to coordinate the release with changes to activity. The current updates to wallets for activity will break the UI.

PR #435, opened 2022-05-03
Rebased search indexer PR + batching

ray-kast:
- Added an AMQP queue config for search indexers.
- Scaffolded a consumer for search index jobs.
- Added foo upsert handler (FIXME) to indexer-search
- Got the search indexer actually running.
- metadata document index
- Refactored search dispatch code into a new module.

### DIFF
--- a/crates/indexer/src/geyser/accounts/name_service.rs
+++ b/crates/indexer/src/geyser/accounts/name_service.rs
@@ -3,20 +3,13 @@ use indexer_core::{
     db::{insert_into, models::TwitterHandle, tables::twitter_handle_name_services, update},
     prelude::*,
 };
-use serde::{Deserialize, Serialize};
 
 use super::Client;
-use crate::prelude::*;
+use crate::{prelude::*, search_dispatch::TwitterHandleDocument};
 
 #[derive(BorshDeserialize, PartialEq, Debug, Clone)]
 struct TwitterHandleAndRegistry {
     registry_key: [u8; 32],
-    handle: String,
-}
-
-#[derive(Deserialize, Serialize)]
-struct TwitterHandleDocument {
-    owner: String,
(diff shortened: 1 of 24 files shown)


crates/indexer/src/http/metadata_json.rs
```diff
@@ -528,3 +546,31 @@ pub async fn process<'a>(
...
+        .and_then(|u| {
+            let id = AssetIdentifier::new(&u);
+
+            proxy_url(client.proxy_args(), &id, Some(("width", "400")))
+                .map(|o| o.map(|u| u.to_string()))
+                .transpose()
+        })
+        .unwrap_or_else(|| Ok(image.to_owned()))?;
+
+    let document = MetadataDocument { image, raw };
```
dandlezzz, line 567: Is this enough data to create a link to the NFT? Wouldn't we need metadata_address?

dandlezzz (changes requested):
Mostly great. Let's talk about what we are indexing and why.

crates/indexer/src/geyser/accounts/name_service.rs
```diff
@@ -26,6 +26,17 @@ pub(crate) async fn process(
 
     let incoming_slot: i64 = slot.try_into()?;
 
+    let document = TwitterHandleDocument {
```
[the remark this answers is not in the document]
ray-kast: I don't think we're gonna be able to do that in a way that's performant or consistent.

crates/indexer/src/http/metadata_json.rs
ray-kast: That's in the `id` field of the document we dispatch for upsert.

dandlezzz (approved):
approved.

PR #452, opened 2022-05-09
metadatajson and twitter handle search queries

imabdulbasit:


kespinola (approved):
hell yeah. 

and in terms of using existing endpoints once this mvp goes through I'd like to move nfts to 100% where we create large doc with all the offers and listings saved on the doc.

crates/graphql/src/schema/query_root.rs
```diff
@@ -441,6 +442,96 @@ impl QueryRoot {
...
+            .await
+            .context("failed to load search result")?
+            .hits;
+
+        debug!("{:?}", result);
+        Ok(result
+            .into_iter()
+            .map(|r| {
+                let value = r.result;
+                MetadataJson {
```
kespinola, line 471: This is a good spot for a from trait.

imabdulbasit: from trait for Serde::value to metadata json?

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -484,3 +484,40 @@ impl NftCount {
...
+        &self.description
+    }
+
+    pub fn category(&self) -> &str {
+        &self.category
+    }
+
+    pub fn attributes(&self) -> Vec<NftAttribute> {
+        self.attributes.clone()
+    }
```
ray-kast, line 522: does `&Vec<...>` as a return type not work?

crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -102,6 +102,10 @@ impl Wallet {
         &self.address
     }
 
+    pub fn twitter_handle(&self) -> Option<String> {
+        self.twitter_handle.clone()
+    }
+
```
ray-kast, line 108: should be `Option<&str>` and use `as_deref` rather than `clone`

crates/graphql/src/schema/query_root.rs
ray-kast: You can deserialize a `Value` into a struct with `serde_json`

crates/graphql/src/schema/query_root.rs
```diff
@@ -441,6 +444,125 @@ impl QueryRoot {
...
+            .context("failed to load search result")?
+            .hits;
+
+        debug!("{:?}", result);
+
+        Ok(result
+            .into_iter()
+            .map(|r| {
+                let value = r.result;
+                Wallet {
```
ray-kast, line 552: This should also just deserialize the value.

crates/indexer/src/search/client.rs
```diff
@@ -44,6 +44,7 @@ impl Client {
     ///
     /// # Errors
     /// This function fails if the Meilisearch database cannot be initialized.
+
```
ray-kast, line 47:
```suggestion
```

crates/graphql/src/schema/query_root.rs
imabdulbasit: yes but we would face the same problem of deserialization like we had with http metadata json indexing.Also, meilisearch query fails even if one of the item can not be deserialized

ray-kast (approved):

PR #453, opened 2022-05-09
Multiple marketplaces query

alchemistgo87:
- marketplaces: Query to fetch list of marketplaces order by name
- Index name column on table store_config_jsons

Screenshot 2022-05-09 at 7 26 43 PM


crates/core/migrations/2022-05-09-133434_index_name_on_store_config_jsons/up.sql
```diff
@@ -0,0 +1,2 @@
+create index if not exists name_store_config_jsons_idx on
+  store_config_jsons using hash (name);
```
kespinola, line 2: I think we want the default index type of btree since we are doing sorts. Hash is good for joining tables on something like an address.

kespinola (changes requested):
Looks great do think we should use btree for the index.

crates/core/migrations/2022-05-09-133434_index_name_on_store_config_jsons/up.sql
alchemistgo87: Ok

ray-kast (changes requested):

kespinola (changes requested):
```graphql
marketplaces(subdomains: ["junglecats", "espi", "monkedao"]) { ... }
```

We should add optionals subdomains filter so we can use on homepage. @kristianeboe can discovery take over this PR?

This will return results in descending order based on subdomain name. Can also reorder based on the subdomain list if provided but I would like not to.

crates/graphql/src/schema/query_root.rs
```diff
@@ -475,6 +475,10 @@ impl QueryRoot {
         let conn = context.shared.db.get()?;
         let mut rows: Vec<models::StoreConfigJson> = store_config_jsons::table
             .filter(store_config_jsons::subdomain.eq(subdomain))
+            .filter(
+                store_config_jsons::store_address
+                    .ne_all(&context.shared.marketplaces_store_address_exclusions),
```
kespinola, line 480:
```suggestion
                    .eq_any(&context.shared.marketplaces_store_address_exclusions),
```
"Filter out any records that are in the provided list".

More idiomatic to use positive assertions.

crates/graphql/src/schema/query_root.rs
```diff
@@ -537,6 +541,51 @@ impl QueryRoot {
...
+        subdomains: Option<Vec<String>>,
+        #[graphql(description = "Limit for query")] limit: Option<i32>,
+        #[graphql(description = "Offset for query")] offset: Option<i32>,
+    ) -> FieldResult<Vec<Marketplace>> {
+        if subdomains.is_none() && limit.is_none() {
+            return Err(FieldError::new(
+                "You must supply limit, subdomains, or both.",
+                graphql_value!({ "Filters": "subdomains: Vec<String>, limit: i32" }),
+            ));
+        }
```
kespinola, line 562: 👍

kespinola (changes requested):
Looks good. Just recommendation on positive assertion.

crates/graphql/src/schema/query_root.rs
austinmilt: Those are opposite operations. `eq_any` will return only those in the provided list. Are you asking for the following (i.e. add `not`)?

```
.filter(
                not(listing_receipts::seller
                    .eq_any(&context.shared.featured_listings_seller_exclusions)),
            )
```

austinmilt: Also, in looking it up, [diesel says](https://docs.diesel.rs/diesel/expression_methods/trait.ExpressionMethods.html#method.ne_all) we should use `ne(all(...))` for postgres dbs

crates/graphql/src/schema/query_root.rs
```diff
@@ -537,6 +542,51 @@ impl QueryRoot {
...
+            .select(store_config_jsons::all_columns)
+            .filter(
+                store_config_jsons::store_address
+                    .ne(all(&context.shared.marketplaces_store_address_exclusions)),
+            )
+            .order(store_config_jsons::name.asc())
+            .into_boxed();
+
+        if let Some(subdomains) = subdomains {
+            query = query.filter(store_config_jsons::subdomain.eq(any(subdomains)));
```
ray-kast, line 576: If `offset` is not used for this branch it should be an error to have `subdomains.is_some() && offset.is_some()`

ray-kast (changes requested):

crates/core/migrations/2022-05-09-133434_index_name_on_store_config_jsons/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table store_config_jsons
+add constraint uniq_subdomain unique (subdomain);
```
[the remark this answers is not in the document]
austinmilt: I havent tested and dont know how to. I could use some help with that.

crates/graphql/src/schema/query_root.rs
austinmilt: Will update to error if there's a weird combination of the 3.

crates/core/migrations/2022-05-09-133434_index_name_on_store_config_jsons/up.sql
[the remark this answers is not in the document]
austinmilt: > Have we tested whether this works on production, or will postgres get mad about conflicts?

this has now been tested on my local machine with a local db

crates/core/migrations/2022-05-31-133434_index_name_on_store_config_jsons/up.sql
```diff
@@ -0,0 +1,8 @@
+delete from store_config_jsons 
+  where store_address='3doAaFs2VuTLnVTPLZwFAWsskqwwC4xLt31dZ24uwYsd';
```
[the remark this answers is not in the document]
kespinola: If the store_config json exists on chain then this delete will only get reverted the next time our `on_startup` validator runs. 

Please just drop and we may need to but a deny list for the marketplaces both for bad actors and bad data.

crates/graphql/src/schema/query_root.rs
kespinola: Thank you sorry about the confusion on my end.

kespinola (approved):

ray-kast (changes requested):
Blocking until we resolve the rows that conflict with the new unique constraint.

ray-kast (changes requested):
[This upsert](https://github.com/holaplex/indexer/blob/c86e04da5afd6a93f04828328b979c65eb346443/crates/indexer/src/http/store_config.rs#L93-L104) needs an `on_conflict(store_config_jsons::subdomain)` clause now.

crates/indexer/src/http/store_config.rs
```diff
@@ -93,12 +93,23 @@ pub async fn process(client: &Client, config_key: Pubkey, uri_str: String) -> Re
...
+            .get_result::<bool>(db);
+
+            if Ok(true) == subdomain_exists {
+                return Ok(())
+            }
+
             insert_into(store_config_jsons::table)
                 .values(&row)
                 .on_conflict(store_config_jsons::config_address)
                 .do_update()
```
kespinola, line 108: I went with a select then exit vs on_conflict because want 2 different on_conflict clauses. If the subdomain is the conflict then do nothing but if address then update but you can't have varying conditions.

kespinola (approved):
@ray-kast can you give it a look again?

ray-kast (approved):

PR #461, opened 2022-05-12
slimmer search document

dandlezzz:
Create a slimmer search document. 

```
pub struct MetadataDocument {
    pub name: String,
    pub mint_address: String,
    pub image: Option<String>,
    pub creator_address: String,
    pub creator_twitter_handle: Option<String>,
}


ray-kast (changes requested):
I'm gonna review how this affects some of the feature flags we use for the search indexer.

ray-kast (approved):
Looks like the removal of MetadataRaw from search_dispatch doesn't affect our dependency tree so all's good there.

imabdulbasit (approved):

PR #464, opened 2022-05-13
Geyser plugin v0.3.1 - Performance Improvements & Instruction Messages

ray-kast:
- Harder, better, faster, hashbrown.
- Turn on transaction notifications.
- Switch to using Tokio for the Geyser plugin.
- Remove default parsing for important config.


mpwsh (approved):
lets go tokioo

PR #465, opened 2022-05-13
holaplex/holaplex/#347: high-res Twitter profile image and bug fix

austinmilt:
# Changes
- adds a high-resolution twitter pfp returned in the `profile` object for `wallet` requests (`profile.rs`)
- fixes query for `follow_wallets` to return wallets not already followed (and not self)

### DIFF
--- a/crates/graphql/src/schema/objects/profile.rs
+++ b/crates/graphql/src/schema/objects/profile.rs
@@ -7,6 +7,7 @@ use super::prelude::*;
 pub struct TwitterProfile {
     pub handle: String,
     pub profile_image_url: String,
+    pub profile_image_url_highres: String,
     pub banner_image_url: String,
     pub description: String,
 }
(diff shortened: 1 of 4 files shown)


crates/graphql/src/schema/objects/profile.rs
```diff
@@ -20,19 +21,17 @@ pub struct Profile {
 }
 
 impl From<TwitterUserProfileResponse> for TwitterProfile {
-    fn from(
-        TwitterUserProfileResponse {
-            screen_name,
-            description,
-            profile_image_url_https,
-            profile_banner_url,
-        }: TwitterUserProfileResponse,
```
kespinola, line 29: We should keep the destructing of the struct.

kespinola (changes requested):

crates/graphql/src/schema/objects/profile.rs
austinmilt: ah, yes, forgot that one. Thanks

ray-kast (approved):

RlozanoR:
Let me get in with y'all

kespinola (approved):

dandlezzz:
> Let me get in with y'all

send us a note, daniel@holaplex.com

PR #471, opened 2022-05-16
Fix-up NFTs Query Accuracy and Improve Performance

kespinola:
### Changes
- Add sea-query as a dependency to the core package.
- Convert the nfts query to use sea-query to form the query but still execute and deserialize with diesel.
- Fix up total counts for listed by filtering out listings that don't belong to the current owner of the metadata. This protects against listing on Holaplex but selling on another exchange. (In the future we should present clean actions to the user to close out dangling listings).

### DIFF
--- a/crates/core/src/db/queries/metadatas.rs
+++ b/crates/core/src/db/queries/metadatas.rs
@@ -16,7 +16,6 @@ use crate::{
         Connection,
     },
     error::prelude::*,
-    prelude::*,
 };
 
 /// Format for incoming filters on attributes
@@ -100,7 +99,6 @@ enum MetadataCollectionKeys {
     Table,
     MetadataAddress,
     CollectionAddress,
-    Verified,
 }
 
 /// List query options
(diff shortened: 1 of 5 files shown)


crates/core/src/db/queries/metadatas.rs
```diff
@@ -85,87 +157,176 @@ pub fn list(
...
+            .equals(Metadatas::Table, Metadatas::Address),
+        );
+
+        query.and_where(
+            Expr::col((
+                MetadataCollectionKeys::Table,
+                MetadataCollectionKeys::CollectionAddress,
+            ))
+            .eq(collection),
+        );
```
kespinola, line 322: @imabdulbasit does this look correct? The caller passes in collection address ie the mint address of the NFT and then inner join the collection address into the the metadata and filter out any that don't match the provided collection address.

kespinola: @dandlezzz @imabdulbasit and I would like to use the mint address of the collection nft as the filter parameter. He mentioned you wanted nft address. Can you share why nft address over mint address?

imabdulbasit: We can use either as filter, shouldn't be a problem but collection mint_address is a field in token metadata.

imabdulbasit (approved):

PR #477, opened 2022-05-17
holaplex/holaplex/349: add optional limit and offset to listings query

austinmilt:
This may be needed for the trending auctions query for the new homepage. Regardless, seems like it cant hurt to add.

### DIFF
--- a/crates/graphql/src/schema/query_root.rs
+++ b/crates/graphql/src/schema/query_root.rs
@@ -355,11 +355,16 @@ impl QueryRoot {
         Ok(Wallet::new(address, twitter_handle))
     }
 
-    fn listings(&self, context: &AppContext) -> FieldResult<Vec<Listing>> {
+    fn listings(
+        &self,
+        context: &AppContext,
+        #[graphql(description = "Query limit")] limit: Option<i32>,
+        #[graphql(description = "Query offset")] offset: Option<i32>,
+    ) -> FieldResult<Vec<Listing>> {
         let now = Local::now().naive_utc();
         let conn = context.shared.db.get()?;
 
-        let rows: Vec<ListingRow> = auction_caches::table
+        let mut query = auction_caches::table
             .inner_join(
                 auction_datas::table.on(auction_caches::auction_data.eq(auction_datas::address)),
             )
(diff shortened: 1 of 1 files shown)


crates/graphql/src/schema/query_root.rs
```diff
@@ -376,8 +381,17 @@ impl QueryRoot {
...
-            .context("Failed to load listings")?;
+            .into_boxed();
+
+        if let Some(limit) = limit {
+            query = query.limit(limit.into());
+        }
+
+        if let Some(offset) = offset {
+            query = query.offset(offset.into());
+        }
```
ray-kast, line 392: Should validate that both or neither limit and offset are provided.

ray-kast (changes requested):

ray-kast (changes requested):
Previous comment about parameter validation still needs addressing.

crates/graphql/src/schema/query_root.rs
```diff
@@ -428,11 +428,41 @@ impl QueryRoot {
...
+        #[graphql(description = "Return only listings whose NFTs' metadatas contain this term")]
+        term: Option<String>,
+        #[graphql(description = "Query limit")] limit: i32,
+        #[graphql(description = "Query offset")] offset: Option<i32>,
+    ) -> FieldResult<Vec<Listing>> {
         let now = Local::now().naive_utc();
-        let conn = context.shared.db.get()?;
+        let conn = context.shared.db.get().context("Failed to connect to DB")?;
+
+        let nft_metadata_addresses: Option<Vec<String>> = match term {
```
austinmilt, line 442: this had to be separated from the additional query filter because rust couldnt pass the query through to the async block this creates.

crates/graphql/src/schema/query_root.rs
```diff
@@ -428,11 +428,46 @@ impl QueryRoot {
...
         let now = Local::now().naive_utc();
-        let conn = context.shared.db.get()?;
+        let conn = context.shared.db.get().context("Failed to connect to DB")?;
+
+        // this had to be separated from the additional query filter because rust 
+        // couldnt pass the query through to the async block this creates.
+        let nft_metadata_addresses: Option<Vec<String>> = match term {
+            // the magic number for limit is just to ensure we can get enough 
+            // addresses that there's a good chance of finding listings that match
+            // with the other query requirements
```
austinmilt, line 447: any ideas how to do this "correctly"?

crates/graphql/src/schema/query_root.rs
austinmilt: I made the limit required and offset optional with default of 0

ray-kast (changes requested):
As discussed on Slack we need to figure out the right way to architect this.

PR #479, opened 2022-05-18
Mpw/moving twitter req to cdn

mpwsh:
Moving twitter requests to assets cdn to allow caching some calls.
This should help us relief the rate limiting we are currently hitting


kespinola (changes requested):
Looks great. Blocking because the change to image_lowres will break clients calling the api.

ray-kast (approved):

PR #486, opened 2022-05-19
Receipts from transactions

imabdulbasit:


crates/indexer/src/geyser/instructions/buy.rs
```diff
@@ -0,0 +1,174 @@
...
+use super::Client;
+use crate::prelude::*;
+
+#[derive(BorshDeserialize, Debug, Clone)]
+pub struct InstructionParameters {
+    trade_state_bump: u8,
+    escrow_payment_bump: u8,
+    buyer_price: u64,
+    token_size: u64,
+}
```
kespinola, line 25: anchor should be exposing this struct as well. not a huge deal but does help keep the source the contract.

https://github.com/metaplex-foundation/metaplex-program-library/blob/master/auction-house/program/src/receipt/mod.rs#L305

crates/indexer/src/geyser/instructions/buy.rs
```diff
@@ -0,0 +1,174 @@
...
+            let offer = select(exists(
+                offers::table.filter(
+                    offers::trade_state
+                        .eq(row.trade_state.clone())
+                        .and(offers::bookkeeper.eq(row.bookkeeper.clone()))
+                        .and(offers::auction_house.eq(row.auction_house.clone()))
+                        .and(offers::buyer.eq(row.buyer.clone()))
+                        .and(offers::metadata.eq(row.metadata.clone()))
+                        .and(offers::price.eq(row.price))
+                        .and(offers::token_size.eq(row.token_size))
```
kespinola, line 105: Again should be covered by trade state address.

If we can boil this down to a few fields also recommend joint index to make the query as fast as possible.

crates/indexer/src/geyser/instructions/cancel.rs
```diff
@@ -0,0 +1,94 @@
...
+                            .eq(row.wallet.clone())
+                            .and(
+                                current_metadata_owners::token_account_address
+                                    .eq(row.token_account.clone()),
+                            ),
+                    ),
+                ))
+                .get_result::<bool>(db);
+
+                if Ok(true) == wallet_is_owner {
```
kespinola, line 61: In general this applies but owners can make offers on their own nfts. Just yesterday degenpoet posted a VOD of making an offer. They made the offer on their own nft.

Thinking we need to query both tables and used the trade_state as the lookup key.

crates/core/migrations/2022-05-19-200746_create_instruction_tables/down.sql
```diff
@@ -0,0 +1,7 @@
+drop table buy_instructions;
+drop table sell_instructions;
+drop table execute_sale_instructions;
+drop table cancel_instructions;
+drop table deposit_instructions;
+drop table withdraw_instructions;
```
kespinola, line 6: In an immediate follow up we should index the withdraw_from_treasury instruction. That will allow us to have a print out of each withdraw from the ah treasury, how much, and to what wallet.

@damianr @dandlezzz this is big for marketplace financials reporting.

kespinola (changes requested):
Some changes to think about. Largest items for me.

- Dropping of unneeded field filters that are covered by trade_state for the existence checks.
- Not using `addresss` as the name for the primary columns of offers, listings, and purchases.

Really really really awesome job.

crates/core/migrations/2022-05-19-200746_create_instruction_tables/up.sql
```diff
@@ -0,0 +1,118 @@
+create table buy_instructions (
+    address                         uuid            primary key default gen_random_uuid(),
```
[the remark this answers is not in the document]
imabdulbasit: yes I'll change this

crates/graphql/src/schema/objects/feed_event.rs
```diff
@@ -117,7 +120,7 @@ pub struct OfferEvent {
     feed_event_id: String,
     twitter_handle: Option<String>,
     wallet_address: String,
-    bid_receipt_address: PublicKey<BidReceipt>,
+    bid_receipt_address: String,
```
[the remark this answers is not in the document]
imabdulbasit: If bid_receipt account is received, I index the account address in feed_events table otherwise the offer uuid if instruction received so bid_receipt_address is changed to String to cast uuid as String. If Pubkey is the type then Uuid when parsed to string would return an error as its not 32 byte

crates/indexer/src/geyser/instructions/buy.rs
imabdulbasit: Oh I missed this so will change it back to anchor one

crates/indexer/src/geyser/instructions/buy.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+async fn upsert_into_offers_table<'a>(
+    client: &Client,
+    data: BuyInstruction<'static>,
+) -> Result<()> {
+    let row = Offer {
+        id: None,
+        trade_state: data.buyer_trade_state,
+        bookkeeper: data.wallet.clone(),
```
kespinola, line 79: I don't think we should track the bookkeeper on the denormalized tables because its only valid for receipts. 

In our does it exsit checks to know if we should override the offers, listings or purchases for the receipt should query the receipts and if it's the first time indexing the receipt (ie receipt doesn't exist in the db) then always override the offer, listing, or purchase.

crates/indexer/src/geyser/instructions/buy.rs
```diff
@@ -0,0 +1,174 @@
...
+        auction_house: data.auction_house,
+        buyer: data.wallet,
+        metadata: data.metadata,
+        token_account: Some(data.token_account),
+        purchase_id: None,
+        price: data.buyer_price,
+        token_size: data.token_size,
+        bump: None,
```
kespinola, line 87: bump can be dropped too. I think this is the bump of the receipt.

crates/core/migrations/2022-05-19-200746_create_instruction_tables/up.sql
```diff
@@ -0,0 +1,118 @@
+create table buy_instructions (
```
kespinola, line 1: @imabdulbasit there are 2 version of the buy instruction.  This is the private buy where the offer is locked to the current owner. What we are using is the publc_buy where the offer is on the NFT and doesn't become void when the nft changes ownership.

We should index both instructions with the public_buy_instruction hopefully being the one used more by other marketplaces.

crates/indexer/src/geyser/instructions/cancel.rs
kespinola: changes to update either the listing or offer depending on the state of nft ownership is good. nice.

kespinola (changes requested):

crates/core/migrations/2022-05-19-200746_create_instruction_tables/up.sql
imabdulbasit: Yes, I know. I'll index the public_buy too

crates/indexer/src/geyser/instructions/buy.rs
imabdulbasit: yes, if the offer is build using the bid_receipt account then bump is not null but I guess that is not important

kespinola (changes requested):
I know its a bit of a backfill nightmare but I think the long term health the system would be better to base feed events solely on listings, offers, and receipts. The associated events tables should link to listing_id, offer_id, or purchase_id instead of the equivalent address.

Also still like split the api changes from the indexing so we can have time to update the frontends.

crates/graphql/src/schema/objects/feed_event.rs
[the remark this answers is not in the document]
imabdulbasit: address or id references?

PR #488, opened 2022-05-19
homepage v2 optimizations

### LINKED ISSUE #485: allow filtering feedEvents query by event types
To facilitate surfacing only the events that are relevant - e.g. on holaplex.com not showing follow events, we should add an optional parameter to the `feedEvents` query to filter returned events by type (e.g. mint, follow, purchase, etc).

austinmilt:
# Changes
- added the ability to exclude certain event types from the feed events query (`feed_event.rs`, `query_root.rs`)
- added a specific query for featured buy-now listings that can exclude some wallets (`query_root.rs`)
- added a temporary hard-coded list for follow wallets (`query_root.rs`)

closes #485


crates/graphql/src/schema/query_root.rs
```diff
@@ -57,6 +63,9 @@ impl QueryRoot {
     ) -> FieldResult<Vec<Wallet>> {
         let conn = ctx.shared.db.get().context("failed to connect to db")?;
 
+        // TODO temporarily force specific wallets
```
austinmilt, line 66: Note this needs to be updated to include only wallets we want to surface _before_ merging. It will affect existing queries from the frontend!

@kespinola @kristianeboe

crates/graphql/src/schema/query_root.rs
```diff
@@ -101,11 +111,31 @@ impl QueryRoot {
         wallet: PublicKey<Wallet>,
         limit: i32,
         offset: i32,
+        exclude_types: Option<Vec<String>>,
     ) -> FieldResult<Vec<FeedEvent>> {
         let conn = ctx.shared.db.get().context("failed to connect to db")?;
 
-        let feed_events =
-            queries::feed_event::list(&conn, wallet, limit.try_into()?, offset.try_into()?)?;
+        let exclusion_types_parsed: Option<Vec<EventType>>;
```
austinmilt, line 118: At some point this should probably work similarly to `PublicKey`, such it is automagically converted from a Graph QL enum to a rust enum, so that the query UI can show valid values.

crates/graphql/src/schema/query_root.rs
```diff
@@ -316,6 +346,41 @@ impl QueryRoot {
...
+        &self,
+        context: &AppContext,
+        #[graphql(description = "Limit for query")] limit: i32,
+        #[graphql(description = "Offset for query")] offset: i32,
+    ) -> FieldResult<Vec<ListingReceipt>> {
+        let conn = context.shared.db.get().context("failed to connect to db")?;
+
+        let holaplex_auction_house_address = "9SvsTjqk3YoicaYnC4VW1f8QAN9ku7QCCk6AyfUdzc9t";
+        // TODO expand with other wallets to exclude
+        let sellers_to_exclude = vec!["7r8oBPs3vNqgqEG8gnyPWUPgWuScxXyUxtmoLd1bg17F"];
```
austinmilt, line 359: This needs to be fleshed out _before_ being merged, so that we dont just return a bunch of listings from holaplex employees.

@kespinola @kristianeboe

ray-kast: This should be changed to load from the runtime environment before merging.  Can discuss ways to make that work.

ray-kast (changes requested):

crates/graphql/src/schema/query_root.rs
austinmilt: nm, I was able to copy it over from the frontend code

crates/graphql/src/schema/query_root.rs
```diff
@@ -101,11 +143,31 @@ impl QueryRoot {
...
+                types
+                    .iter()
+                    .map(|v| EventType::from_str(v))
+                    .filter(|v| !v.is_err())
+                    .map(Result::unwrap)
+                    .collect(),
+            );
+        } else {
+            exclusion_types_parsed = None;
+        }
```
kespinola, line 162: You can use .map on excluded_types which will return none if its none or transform the value. If you want to rewrap in an Option then and_then will be what you want.

ray-kast (approved):

PR #500, opened 2022-05-25
#499: change featured_listings to choose NFTs whose creators have many followers

### LINKED ISSUE #499: Move "what's hot" query to prioritize nfts with much followed creators

austinmilt:
This PR changes the `featured_listings` query to select listings whose creators have many followers by sorting results by the sum of followers of NFT creators.

To increase query speed, the PR also adds filters to only consider creators who have > 0 share and are verified.

closes #499

### DIFF
--- a/crates/core/src/db/queries/featured_listings.rs
+++ b/crates/core/src/db/queries/featured_listings.rs
@@ -14,25 +14,25 @@ use crate::{
 };
 
 const FEATURED_LISTINGS_QUERY: &str = r"
-SELECT 
-    a.address, 
-    a.trade_state, 
-    a.bookkeeper, 
-    a.auction_house, 
-    a.seller, 
-    a.metadata, 
-    a.purchase_receipt, 
-    a.price, 
-    a.token_size, 
-    a.bump, 
+SELECT
+    a.address,
+    a.trade_state,
+    a.bookkeeper,
+    a.auction_house,
+    a.seller,
+    a.metadata,
+    a.purchase_receipt,
+    a.price,
+    a.token_size,
+    a.bump,
     a.trade_state_bump,
-    a.created_at, 
+    a.created_at,
     a.canceled_at
-    
+
 FROM (
 
-    SELECT 
-        listing_receipts.*, 
+    SELECT
+        listing_receipts.*,
         row_number() OVER (
             PARTITION BY listing_receipts.seller ORDER BY listing_receipts.price DESC
         ) as row
--- a/crates/core/src/db/queries/featured_listings.rs
+++ b/crates/core/src/db/queries/featured_listings.rs
@@ -1,0 +1,66 @@
+//! Query utilities for featured listings.
+
+use anyhow::Context;
+use diesel::{
+    pg::Pg,
+    sql_types::{Array, Integer, Nullable, Text},
+    types::ToSql,
+    RunQueryDsl,
+};
+
+use crate::{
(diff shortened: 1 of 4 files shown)


austinmilt:
converted to draft because we risk featuring only listings from the top profile if they have many listings. There may be a solution to limit the number of listings per seller https://stackoverflow.com/a/49049105

crates/graphql/src/schema/query_root.rs
```diff
@@ -358,8 +366,11 @@ impl QueryRoot {
                 listing_receipts::seller
                     .ne_all(&context.shared.featured_listings_seller_exclusions),
             )
-            .filter(listing_receipts::seller.eq(wallet_totals::address))
-            .order(wallet_totals::followers.desc())
+            .filter(metadata_creators::share.gt(0))
+            .filter(metadata_creators::verified.eq(true))
+            .group_by(listing_receipts::metadata)
+            .group_by(listing_receipts::address)
```
ray-kast, line 372: `group_by` is a weird undocumented diesel function.  did you mean `distinct_on`?

ray-kast (changes requested):

crates/graphql/src/schema/query_root.rs
austinmilt: > `group_by` is a weird undocumented diesel function. did you mean `distinct_on`?

no, I am using an aggregator `sum`, which requires `group_by` to work because it allows aggregation, whereas `distinct` would only return a single row for each and bypass the sum

_I think_

crates/core/src/db/queries/featured_listings.rs
```diff
@@ -0,0 +1,66 @@
...
+};
+
+use crate::{
+    db::{models::ListingReceipt, Connection},
+    error::Result,
+};
+
+const FEATURED_LISTINGS_QUERY: &str = r"
+SELECT a.address, a.trade_state, a.bookkeeper, a.auction_house, a.seller, a.metadata, a.purchase_receipt, a.price, a.token_size, a.bump, a.trade_state_bump, a.created_at, a.canceled_at from (
+    SELECT listing_receipts.*, row_number() OVER (PARTITION BY listing_receipts.seller ORDER BY listing_receipts.price DESC) as row
```
ray-kast, line 18: Looks like this exceeds the formatter's 100-char line limit, can we get some newlines?

ray-kast (changes requested):

crates/graphql/src/schema/query_root.rs
```diff
@@ -338,32 +338,33 @@ impl QueryRoot {
     fn featured_listings(
         &self,
         context: &AppContext,
+        #[graphql(description = "Return listings only from these auction houses")]
+        auction_houses: Option<Vec<String>>,
+        #[graphql(description = "Return listings not from these sellers")]
+        seller_exclusions: Option<Vec<String>>,
```
kespinola, line 344: The type for PublicKey implements ToSQL so should be able to reference the arguments as public keys.

```suggestion
        auction_houses: Option<Vec<PublicKey<AuctionHouse>>>,
        #[graphql(description = "Return listings not from these sellers")]
        seller_exclusions: Option<Vec<PublicKey<Wallet>>>,
```

kespinola (changes requested):
Nice. Only item is the comment on public key for arguments instead of string.

kespinola (approved):

ray-kast (approved):

PR #506, opened 2022-06-02
Auction house instructions indexing

imabdulbasit:


crates/indexer/src/geyser/instructions/buy.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    client
+        .db()
+        .run(move |db| {
+            insert_into(buy_instructions::table)
+                .values(&row)
+                .execute(db)
+        })
+        .await
+        .context("failed to insert buy instruction ")?;
```
kespinola, line 54: In future move this to a transaction?

crates/indexer/src/geyser/instructions/cancel.rs
```diff
@@ -0,0 +1,83 @@
+use borsh::BorshDeserialize;
+use indexer_core::db::{
+    insert_into,
+    models::CancelInstruction,
+    select,
+    tables::{cancel_instructions, listings, offers},
+    update,
+};
+use mpl_auction_house::instruction::Cancel;
```
kespinola, line 9: Nice getting the cancel struct from the program 👍

kespinola (approved):
Great work.

ray-kast (changes requested):
Are there any tables we need to add the `check_slot_wv` trigger to?

ray-kast (changes requested):
Looks good.  Still gonna advocate for the use of `check_slot_wv` - we can use `-1` as a fallback for any instructions that don't have these fields.

ray-kast (approved):

PR #509, opened 2022-06-06
[GQL] Added `update_authority_address` to the NFT Schema

roederw:
`update_authority_address` isn't exposed via the NFT object, but is accessible via the `metadatas` table. Adding the parameter to the schema to be used directly from the NFT.


kespinola (approved):

kespinola:
This is releasing to our dev we'll test it out there and release to prod.

PR #510, opened 2022-06-06
holaplex #348: add filter for nfts for those with active offers

austinmilt:
This PR adds an option filter to `nfts` to return only those with active offers.

# Changes
- added `with_offers` filter to `nfts` query (`metadatas.rs`, `query_root.rs`)
- fixed a bug where `update_authority_address` wasnt being queried from SQL, so the `Nft` object would fail to be created from the rows returned from the query (`metadatas.rs`)


crates/core/src/db/queries/metadatas.rs
```diff
@@ -264,23 +267,36 @@ pub fn list(
...
+                .add(Expr::col((BidReceipts::Table, BidReceipts::Buyer)).is_in(offerers))
+                .add(Expr::tbl(BidReceipts::Table, BidReceipts::PurchaseReceipt).is_null())
+                .add(Expr::tbl(BidReceipts::Table, BidReceipts::CanceledAt).is_null());
+        } else if let Some(with_offers) = with_offers {
+            if with_offers {
+                bid_receipts_conditions = bid_receipts_conditions
+                    .add(Expr::tbl(BidReceipts::Table, BidReceipts::PurchaseReceipt).is_null())
+                    .add(Expr::tbl(BidReceipts::Table, BidReceipts::CanceledAt).is_null());
+            }
+        }
```
kespinola, line 291: Instead of doing special logic for one or the other with_offers should tack on filters to bid receipt where canceled_at isn't null but keep the offerers logic. If someone calls the query with_offers and offerers then they should get the nfts for the addresses set as offerers. If they just set with_offers should pick any nfts with an offer for the supplied auction_houses.

kespinola (changes requested):

crates/core/src/db/queries/metadatas.rs
austinmilt: Updated to remove the `else`, which I think is what you were asking for. 

Also, this PR is still a draft because I am getting a runtime error  I dont know how to track down:

> Caused by:
>     Column `update_authority_address` was not present in query

crates/core/src/db/queries/metadatas.rs
```diff
@@ -264,23 +269,38 @@ pub fn list(
...
-    if let Some(offerers) = offerers {
+    // ignoring the with_offers=false case for now, since it is more complicated and we dont currently have a need
+    let should_query_bid_receipts =
+        offerers.is_some() || (with_offers.is_some() && with_offers.unwrap_or(false));
+
+    if should_query_bid_receipts {
+        let mut bid_receipts_conditions = Condition::all().add(
+            Expr::tbl(BidReceipts::Table, BidReceipts::Metadata)
+                .equals(Metadatas::Table, Metadatas::Address),
+        );
```
kespinola, line 280: Isn't this always needed since its referenced on LN303?

austinmilt: line 303 is under the should_query_bid_receipts block, so it will always be created when it is going to be used.

crates/core/src/db/queries/metadatas.rs
austinmilt: The reported bug is fixed in this PR

crates/core/src/db/queries/metadatas.rs
```diff
@@ -264,23 +269,38 @@ pub fn list(
...
+
+        if let Some(offerers) = offerers {
+            bid_receipts_conditions = bid_receipts_conditions
+                .add(Expr::col((BidReceipts::Table, BidReceipts::Buyer)).is_in(offerers))
+                .add(Expr::tbl(BidReceipts::Table, BidReceipts::PurchaseReceipt).is_null())
+                .add(Expr::tbl(BidReceipts::Table, BidReceipts::CanceledAt).is_null());
+        }
+
+        if let Some(with_offers) = with_offers {
+            if with_offers {
```
ray-kast, line 290: Can't this just be `if let Some(true) = with_offers`?  If so then why's it an option?

crates/core/src/db/queries/metadatas.rs
```diff
@@ -264,23 +269,38 @@ pub fn list(
         );
     }
 
-    if let Some(offerers) = offerers {
+    // ignoring the with_offers=false case for now, since it is more complicated and we dont currently have a need
+    let should_query_bid_receipts =
+        offerers.is_some() || (with_offers.is_some() && with_offers.unwrap_or(false));
+
+    if should_query_bid_receipts {
```
ray-kast, line 276: Unnecessary let-binding?

ray-kast (changes requested):

crates/core/src/db/queries/metadatas.rs
austinmilt: I didnt know about the `if let Some(true)` unpacking, pretty cool.

It's an option because `with_offers = false` implies you only want listings _without offers_ and that's not something I want to try to support right now. So, if you set it to `false` or dont provide it, then we ignore whether or not the NFT has offers, i.e. you'll get back NFTs with offers and without offers (versus only those with offers when `with_offers = true`)

crates/core/src/db/queries/metadatas.rs
ray-kast: Yes please.

crates/core/src/db/queries/metadatas.rs
ray-kast: Hmm.  Should make a note or an error in the code for that if we don't support it, otherwise the API surface becomes confusing.

austinmilt: It's already noted in the API documentation, but I will throw an error if it is `false` just to drive home the point :D

crates/graphql/src/schema/query_root.rs
```diff
@@ -315,6 +319,13 @@ impl QueryRoot {
             ));
         }
 
+        if with_offers.is_some() && !with_offers.unwrap() {
```
ray-kast, line 322:
```suggestion
        if let Some(false) = with_offers {
```

ray-kast (changes requested):

crates/core/src/db/queries/metadatas.rs
```diff
@@ -264,23 +269,32 @@ pub fn list(
         );
     }
 
-    if let Some(offerers) = offerers {
+    if offerers.is_some() || (with_offers.is_some() && with_offers.unwrap_or(false)) {
```
kespinola, line 272: with_offers shouldn't be dependent on offerers. The offerers filter is to find  offers on an nfts for a set of wallets. The with_offers should find any nft that has at least 1 offer for the provided auction houses.

I'd like to see 1 block for offers and 1 for with_offers instead of having them handled in the same block.

kespinola, line 272:
```suggestion
    if offerers.is_some() || with_offers.is_some() {
```
Since LN285 only matches on true conditions logic is easier to follow with just the OR condition.

https://github.com/holaplex/indexer/pull/510/files#diff-9a080dfe669d746b43c043fc557a06d5add886b23833af6daa7d80745c432520R285

kespinola (changes requested):

crates/core/src/db/queries/metadatas.rs
austinmilt: With your change, if `offerers` is `None` and `with_offers` is `false`, then this block will execute, adding query conditions that are not actually used for anything. You could argue that since the current version of the query will fail if `with_offers` is `false`, then that condition could not be met. However, that seems like an antipattern to me, to make an assumption about the value of `with_offers` based on the code reaching this point.

kespinola: block the execution? The query doesnt fail if with_offers is false it just removes nfts that don't have offers which isn't the normal usage pattern but is a valid query condition.

kespinola: Ok is see only matching on true then unwrap with_offers before and then references bellow.

```rust
let with_offers = with_offers.unwrap_or(false); // not that user can still pass false to the option which will just skip adding the filter bc its false

if offerers.is_some() || with_offers {
  ...
  
  if with_offers {
    // add extract bid filter
  }
}
```

kespinola (changes requested):

kespinola (approved):

Issue #514, opened 2022-06-09
New GQL endpoint for requesting multiple walletProfiles at a time


kristianeboe:
In the new messages page I end up with a potentially long list of pubkeys/addresses that have sent me a message. I could just do an api call to the "wallet" query endpoint pr address, but I would rather give the whole list and get a list of profiles back. If the address does not have twitter attached to it, just return the address.

Everything must be "in-order" :) 

Something like
```
{
  wallet(s)?(addresses:[
    "aaaa",
    "bbbb"
  ]) {
    address
    profile {
      handle
      profileImageUrlLowres
    }
  }
}
```

PR #515, opened 2022-06-09
#514: add query for multiple wallets

austinmilt:
This PR adds a list version of the `wallet` query.

# Changes
- added method to get multiple twitter handles from a vector of wallet addresses (`twitter_handle_name_services.rs`, `models.rs`). Note that this uses the `sql_query` method from diesel because of the need to select the row with the highest write version.
- used the new query above to get the handles for all provided wallet addresses and then turned these into wallets (`query_root.rs`)

closes #515

### DIFF
--- a/crates/core/src/db/queries/twitter_handle_name_service.rs
+++ b/crates/core/src/db/queries/twitter_handle_name_service.rs
@@ -4,12 +4,13 @@ use diesel::{
     expression::{AsExpression, NonAggregate},
     pg::Pg,
     query_builder::{QueryFragment, QueryId},
-    sql_types::Text,
+    sql_types::{Array, Text},
+    types::ToSql,
     AppearsOnTable, OptionalExtension,
 };
 
 use crate::{
-    db::{tables::twitter_handle_name_services, Connection},
+    db::{models::TwitterHandle, tables::twitter_handle_name_services, Connection},
     error::Result,
     prelude::*,
 };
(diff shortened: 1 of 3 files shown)


kespinola (dismissed):

crates/core/src/db/queries/twitter_handle_name_service.rs
```diff
@@ -32,3 +33,31 @@ where
...
+    addresses: impl ToSql<Array<Text>, Pg>,
+) -> Result<Vec<TwitterHandle>> {
+    // Appears to not be possible to select a row by unique key by the
+    // highest value in another row using diesel ORM
+    // See https://stackoverflow.com/a/61964064 (and diesel gitter)
+    diesel::sql_query(GET_MULTIPLE_HANDLES_QUERY)
+        .bind(addresses)
+        .load(conn)
+        .context("Failed to load featured listings")
+}
```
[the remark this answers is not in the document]
austinmilt: Happy to convert it to simple diesel query. This was in response to the issues we were having with wrong twitter handle being served, or serving an outdated handle.

crates/core/src/db/queries/twitter_handle_name_service.rs
```diff
@@ -32,3 +33,31 @@ where
...
+
+const GET_MULTIPLE_HANDLES_QUERY: &str = r"
+SELECT A.*
+FROM twitter_handle_name_services A
+INNER JOIN (
+    SELECT wallet_address, MAX(write_version) as write_version
+    FROM twitter_handle_name_services
+    WHERE wallet_address = ANY($1)
+    GROUP BY wallet_address
+) B on A.wallet_address = B.wallet_address AND A.write_version = B.write_version
```
ray-kast, line 45: Why are we querying write version?

ray-kast (changes requested):

crates/core/src/db/queries/twitter_handle_name_service.rs
austinmilt: Misguided attempt to avoid returning multiple entries for a single wallet.

crates/core/src/db/queries/twitter_handle_name_service.rs
```diff
@@ -32,3 +32,15 @@ where
...
+/// Return twitter handles linked to the provide wallet addresses
+///
+/// # Errors
+/// This function fails if the underlying query fails to execute.
+pub fn get_multiple(conn: &Connection, addresses: Vec<String>) -> Result<Vec<TwitterHandle>> {
+    twitter_handle_name_services::table
+        .filter(twitter_handle_name_services::wallet_address.eq(any(addresses)))
+        .select(twitter_handle_name_services::all_columns)
+        .load(conn)
+        .context("Failed to load twitter handle")
```
ray-kast:
```suggestion
        .context("Failed to load twitter handles")
```

ray-kast (approved):

PR #517, opened 2022-06-10
holaplex/#348: nfts search term

austinmilt:
This PR adds an optional search term to the `nfts` query. It works by getting a list of metadata addresses from meili search matching the search term, and passing those addresses into the SQL query.

# Changes
- added optional metadata `addresses` to the `metadatas::list` query (`metadatas.rs`, `query_root.rs`). This is treated the same way as other filters, i.e. it returns NFTs whose metadata addresses match those, but it may further refine the results when combined with other filters.
- added optional search term `term` to the `nfts` query, which passes metadata addresses returned by meili to `metadatas::list` (`query_root.rs`)
- also contains the `update_authority_address` bugfix (`metadatas.rs`)

### DIFF
--- a/crates/graphql/src/schema/query_root.rs
+++ b/crates/graphql/src/schema/query_root.rs
@@ -4,7 +4,6 @@ use indexer_core::{
         queries::{self, feed_event::EventType},
         tables::twitter_handle_name_services,
     },
-    meilisearch::client::Client,
 };
 use objects::{
     auction_house::AuctionHouse,
--- a/crates/graphql/src/schema/query_root.rs
+++ b/crates/graphql/src/schema/query_root.rs
@@ -1,9 +1,7 @@
-use indexer_core::{
-    db::{
-        expression::dsl::all,
-        queries::{self, feed_event::EventType},
-        tables::twitter_handle_name_services,
-    },
+use indexer_core::db::{
+    expression::dsl::all,
+    queries::{self, feed_event::EventType},
+    tables::twitter_handle_name_services,
 };
 use objects::{
     auction_house::AuctionHouse,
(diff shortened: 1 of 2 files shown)


crates/graphql/src/schema/query_root.rs
```diff
@@ -308,16 +312,42 @@ impl QueryRoot {
...
+
+                Some(
+                    search_result
+                        .into_iter()
+                        .map(|r| MetadataJson::from(r.result).address)
+                        .collect(),
+                )
+            },
+            None => None,
+        };
```
kespinola, line 347: You can use a map on the optional term to run the transform when the term has some value.

```suggestion
        let search_addresses = term.map(|term| {
                 let search = &context.shared.search;
                let search_result = search
                    .index("metadatas")
                    .search()
                    .with_query(&term)
                    .with_offset(offset.try_into()?)
                    .with_limit(limit.try_into()?)
                    .execute::<Value>()
                    .await
                    .context("failed to load search result for metadata json")?
                    .hits;
        })
```

kespinola (changes requested):
Just the map refactor. Nice job!

crates/graphql/src/schema/query_root.rs
austinmilt: Do you know an elegant way to use `Option::map` with an async operation? [My googling](https://users.rust-lang.org/t/how-to-run-async-code-within-option-result-function-chain/64053) makes it seem like you have to extract the operation to a separate function and then use that function. In that case, what I currently have is clearer/cleaner.

kespinola: Ah the async nvm then.

kespinola (approved):

PR #523, opened 2022-06-14
add collection nft data loader

milorue:
Adds any related collection nfts to the Nft Object
```
nft{
  ...
  collections: [Nft]
}
```

Allows us to show collection info for any given nft in support of holaplex.com's collection support


ray-kast (changes requested):
This doesn't seem to work.  Not sure why or if we simply don't have enough data.

ray-kast (approved):

PR #531, opened 2022-06-15
holaplex/#541: add stats for profiles

austinmilt:
This PR adds an endpoint to serve stats across profiles. Right now that's just the total number of profiles, but will be updated in the future with a breakdown of different types of profiles (e.g. creators, collectors, etc).

### DIFF
--- a/crates/graphql/src/schema/objects/profile.rs
+++ b/crates/graphql/src/schema/objects/profile.rs
@@ -52,3 +52,21 @@ impl TwitterProfile {
         &self.description
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct ProfilesStats {}
+
+#[graphql_object(Context = AppContext)]
+impl ProfilesStats {
+    #[graphql(description = "The total number of indexed profiles")]
+    fn total_profiles(&self, context: &AppContext) -> FieldResult<i32> {
+        let conn = context.shared.db.get()?;
+
+        let count: i64 = twitter_handle_name_services::table
+            .count()
+            .get_result(&conn)
+            .context("failed to load total profiles count")?;
+
+        Ok(count.try_into()?)
+    }
+}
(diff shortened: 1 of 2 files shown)


crates/graphql/src/schema/objects/profile.rs
```diff
@@ -52,3 +52,21 @@ impl TwitterProfile {
         &self.description
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct ProfilesStats {}
```
ray-kast, line 57:
```suggestion
pub struct ProfilesStats;
```

crates/graphql/src/schema/query_root.rs
```diff
@@ -585,6 +585,11 @@ impl QueryRoot {
             .collect::<Vec<Wallet>>())
     }
 
+    #[graphql(description = "returns stats about profiles")]
+    async fn profiles_stats(&self) -> FieldResult<ProfilesStats> {
+        Ok(ProfilesStats {})
```
ray-kast, line 590:
```suggestion
        Ok(ProfilesStats)
```

ray-kast (changes requested):
You should be able to use a unit struct rather than an empty named field struct.

crates/graphql/src/schema/objects/profile.rs
```diff
@@ -52,3 +52,21 @@ impl TwitterProfile {
         &self.description
     }
 }
+
+#[derive(Debug, Clone)]
```
ray-kast, line 56:
```suggestion
#[derive(Debug, Clone, Copy)]
```

crates/graphql/src/schema/query_root.rs
```diff
@@ -585,6 +585,11 @@ impl QueryRoot {
             .collect::<Vec<Wallet>>())
     }
 
+    #[graphql(description = "returns stats about profiles")]
+    async fn profiles_stats(&self) -> FieldResult<ProfilesStats> {
+        Ok(ProfilesStats)
+    }
```
ray-kast, line 591:
```suggestion
    #[graphql(description = "returns stats about profiles")]
    async fn profiles_stats(&self) -> ProfilesStats {
        ProfilesStats
    }
```

ray-kast (changes requested):
Forgot a couple revisions.

PR #539, opened 2022-06-21
Release 6/21/2022

kespinola:


kespinola (commented):
Blocking too. Collections should be Option<Nft> instead of Vec<CollectionNft>. 

Looking into this with @milorue today.

mpwsh (approved):

PR #540, opened 2022-06-22
NFT collection field returns single item

kespinola:
### Changes
- Switch the collection loader to return `Option<CollectionNft>` since an NFT can only be assigned to 1 collection at a time.


austinmilt (approved):

milorue (approved):
Awesome looks good

PR #541, opened 2022-06-22
(Draft) #534: add featured collections queries

austinmilt:
WIP

### DIFF
--- a/crates/core/src/db/queries/collections.rs
+++ b/crates/core/src/db/queries/collections.rs
@@ -1,0 +1,61 @@
+//! Query utilities for collections.
+
+use anyhow::Context;
+use diesel::{pg::Pg, sql_types::{Integer, Text, Nullable, Array}, types::ToSql, RunQueryDsl};
+
+use crate::{
+    db::{custom_types::OrderDirection, models::Nft, Connection},
+    error::Result,
+};
+
+const COLLECTIONS_BY_VOLUME_QUERY: &str = r"
+select
+    metadatas.address,
+    metadatas.name,
+    metadatas.seller_fee_basis_points,
+    metadatas.update_authority_address,
+    metadatas.mint_address,
+    metadatas.primary_sale_happened,
+    metadatas.uri,
+    metadatas.slot,
+    metadata_jsons.description,
+    metadata_jsons.image,
+    metadata_jsons.category,
+    metadata_jsons.model
+from metadata_jsons
+inner join metadatas on (metadatas.address = metadata_jsons.metadata_address)
+inner join (
+    select metadata_collection_keys.collection_address as collection, sum(purchase_receipts.price) as volume
+        from purchase_receipts
+        inner join metadatas on (purchase_receipts.metadata = metadatas.address)
+        inner join metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
+        where ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
+        group by metadata_collection_keys.collection_address
+        order by volume {order_direction}
+		limit $2
+		offset $3
+) a on (a.collection = metadatas.mint_address)
+-- $2: limit::text[]
+-- $2: limit::integer
+-- $3: offset::integer";
+
+/// Query collections ordered by volume
+///
+/// # Errors
+/// returns an error when the underlying queries throw an error
+pub fn by_volume(
+    conn: &Connection,
+    addresses: impl ToSql<Nullable<Array<Text>>, Pg>,
+    order_direction: OrderDirection,
+    limit: impl ToSql<Integer, Pg>,
+    offset: impl ToSql<Integer, Pg>,
+) -> Result<Vec<Nft>> {
+    let query =
+        COLLECTIONS_BY_VOLUME_QUERY.replace("{order_direction}", &order_direction.to_string());
+    diesel::sql_query(query)
+        .bind(addresses)
+        .bind(limit)
+        .bind(offset)
+        .load(conn)
+        .context("Failed to load collections by volume")
+}
--- a/crates/graphql/src/schema/query_root.rs
+++ b/crates/graphql/src/schema/query_root.rs
@@ -1,4 +1,5 @@
 use indexer_core::db::{
+    self,
     expression::dsl::all,
     queries::{self, feed_event::EventType},
     tables::twitter_handle_name_services,
@@ -40,6 +41,24 @@ struct AttributeFilter {
 impl From<AttributeFilter> for queries::metadatas::AttributeFilter {
     fn from(AttributeFilter { trait_type, values }: AttributeFilter) -> Self {
         Self { trait_type, values }
+    }
+}
+
+#[derive(Debug, Clone, Copy, juniper::GraphQLEnum)]
+#[graphql(description = "Sorts results ascending or descending")]
+pub enum OrderDirection {
+    #[graphql(name = "DESC")]
+    Desc,
+    #[graphql(name = "ASC")]
+    Asc,
+}
+
+impl From<OrderDirection> for db::custom_types::OrderDirection {
+    fn from(other: OrderDirection) -> Self {
+        match other {
+            OrderDirection::Desc => Self::Desc,
+            OrderDirection::Asc => Self::Asc,
+        }
     }
 }
 
@@ -128,6 +147,56 @@ impl QueryRoot {
         )?;
 
         feed_events
+            .into_iter()
+            .map(TryInto::try_into)
+            .collect::<Result<_, _>>()
+            .map_err(Into::into)
+    }
+
+    #[graphql(description = "Returns featured collection NFTs ordered by volume (sum over prices)")]
+    async fn collections_featured_by_volume(
+        &self,
+        context: &AppContext,
+        #[graphql(term = "Return collections whose name or family contain this term (case insensitive)")] term: Option<
+            String,
+        >,
+        #[graphql(order_direction = "Sort ascending or descending")]
+        order_direction: OrderDirection,
+        #[graphql(limit = "Return at most this many results")] limit: i32,
+        #[graphql(offset = "Return results starting from this index")] offset: i32,
+    ) -> FieldResult<Vec<Nft>> {
+        let conn = context.shared.db.get().context("failed to connect to db")?;
+
+        let addresses: Option<Vec<String>> = match term {
+            Some(term) => {
+                let search = &context.shared.search;
+                let search_result = search
+                    //TODO refactor for collections
+                    .index("metadatas")
+                    .search()
+                    .with_query(&term)
+                    .with_offset(offset.try_into()?)
+                    .with_limit(limit.try_into()?)
+                    .execute::<Value>()
+                    .await
+                    .context("failed to load search result for collections")?
+                    .hits;
+
+                Some(
+                    search_result
+                        .into_iter()
+                        // TODO refactor for collections
+                        .map(|r| MetadataJson::from(r.result).address)
+                        .collect(),
+                )
+            },
+            None => None,
+        };
+        
+        let collections =
+            queries::collections::by_volume(&conn, addresses, order_direction.into(), limit, offset).unwrap();
+
+        collections
             .into_iter()
             .map(TryInto::try_into)
             .collect::<Result<_, _>>()
--- a/crates/core/src/db/custom_types.rs
+++ b/crates/core/src/db/custom_types.rs
@@ -185,3 +185,15 @@ impl FromSql<ListingEventLifecycle, Pg> for ListingEventLifecycleEnum {
         from_bytes(bytes)
     }
 }
+
+/// Direction for ordering SQL query results by the "ORDER BY" variable(s)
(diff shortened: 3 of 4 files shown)


crates/core/src/db/queries/collections.rs
```diff
@@ -0,0 +1,66 @@
...
+/// returns an error when the underlying queries throw an error
+pub fn by_volume(
+    conn: &Connection,
+    addresses: impl ToSql<Nullable<Array<Text>>, Pg>,
+    order_direction: OrderDirection,
+    limit: impl ToSql<Integer, Pg>,
+    offset: impl ToSql<Integer, Pg>,
+) -> Result<Vec<Nft>> {
+    let query =
+        COLLECTIONS_BY_VOLUME_QUERY.replace("{order_direction}", &order_direction.to_string());
```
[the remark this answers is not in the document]
austinmilt: done!

crates/core/src/db/queries/collections.rs
```diff
@@ -0,0 +1,69 @@
...
+            WHERE ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
+            GROUP BY metadata_collection_keys.collection_address
+            ORDER BY volume {order_direction}
+            LIMIT $2
+            OFFSET $3
+    ) a ON (a.collection = metadatas.mint_address)
+    -- $1: addresses::text[]
+    -- $2: limit::integer
+    -- $3: offset::integer",
+        order_direction = order_direction
```
kespinola, line 67: Does order_direction not work as a sql parameter?

crates/core/src/db/queries/collections.rs
```diff
@@ -0,0 +1,69 @@
...
+    INNER JOIN (
+        SELECT metadata_collection_keys.collection_address AS collection, SUM(purchase_receipts.price) AS volume
+            FROM purchase_receipts
+            INNER JOIN metadatas on (purchase_receipts.metadata = metadatas.address)
+            INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
+            WHERE ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
+            GROUP BY metadata_collection_keys.collection_address
+            ORDER BY volume {order_direction}
+            LIMIT $2
+            OFFSET $3
```
kespinola, line 62: Instead of re-running this query on each row query performance may be improved if assembly of the volume stats is done in its own query then references using WITH clause.

kespinola (changes requested):
This is looking good. 1 query adjustment for discussion and question on parameterization of the query.

crates/core/src/db/queries/collections.rs
austinmilt: Looks to improve the query by up to ~200 ms 🎉

crates/core/src/db/queries/collections.rs
austinmilt: Short answer is not well. I tried to do so originally, but it's not as simple as other variables and caused queries to fail. The only [solutions I could find](https://stackoverflow.com/questions/3884884/conditional-sql-order-by-asc-desc-for-alpha-columns) that allowed a variable order direction in the prepared statement was to use some pretty ugly SQL features, whereas string formatting is straightforward.

crates/core/src/db/queries/collections.rs
```diff
@@ -0,0 +1,70 @@
...
+        .bind(offset)
+        .load(conn)
+        .context("Failed to load collections by volume")
+}
+
+fn make_by_volume_query_string(order_direction: OrderDirection) -> String {
+    format!(
+        r"
+        WITH volume_table (collection, volume) as (
+            SELECT metadata_collection_keys.collection_address AS collection, SUM(purchase_receipts.price) AS volume
```
imabdulbasit, line 39: purchases table instead of purchase_receipts to be used here so that we can get collections for marketplaces which doesn't use receipts

imabdulbasit (dismissed):

crates/graphql/src/schema/query_root.rs
```diff
@@ -694,6 +713,62 @@ impl QueryRoot {
...
+            },
+            None => None,
+        };
+
+        let collections = queries::collections::by_volume(
+            &conn,
+            addresses,
+            order_direction.into(),
+            limit,
+            offset,
```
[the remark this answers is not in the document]
kespinola: Probably best to cap the search the results by a fixed number say 100. Or to start no limit since final result chunked by the sql query.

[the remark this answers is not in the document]
austinmilt: @ray-kast Unfortunately there's no way around having unexpected pagination behavior as long as meilisearch happens _before_ the SQL query (and meili results are potentially unbounded). There was a discussion between myself and Mariano in slack to this effect a week or two ago. It's just simply the wrong ordering of operations. See comment above for how I'll try to reduce the impact for now.

kespinola (approved):

crates/graphql/src/schema/query_root.rs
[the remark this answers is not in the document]
kespinola: >I’m happy with how it is currently. For the user best search experience is through meli. It may cut total results on single letter searches like a, f, etc but real unlikely for user to scroll through a 1000 search results.

https://holaplex.slack.com/archives/C0339RA5M3P/p1657043838973919

Discussion in slack.

crates/core/src/db/queries/collections.rs
```diff
@@ -0,0 +1,80 @@
...
+        r"
+        WITH volume_table (collection, volume) as (
+            SELECT metadata_collection_keys.collection_address AS collection, SUM(purchases.price) AS volume
+                FROM purchases
+                INNER JOIN metadatas on (purchases.metadata = metadatas.address)
+                INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
+                WHERE
+                    ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
+                    AND purchases.created_at >= $2
+                    AND purchases.created_at <= $3
```
kespinola, line 51: Same applies on this query where just want to do volume for SOL based auction houses.

```sql
INNER JOIN auction_houses on (listings.auction_house = auction_houses.address)
  WHERE
    ...
    AND auction_houses.treasury_mint = 'So11111111111111111111111111111111111111112'

**Review by kespinola — CHANGES_REQUESTED** [review]
Sorry. just reviewed the other PR. reminded me about filtering for SOL based auction houses.

File: crates/core/src/db/queries/collections.rs (hunk 6 above)
Comment (austinmilt) [reply crates/core/src/db/queries/collections.rs:53]: done!

File: crates/core/src/db/queries/collections.rs [hunk 7]
```diff
@@ -0,0 +1,149 @@
...
+            SELECT metadata_collection_keys.collection_address AS collection, MIN(listings.price) * COUNT( DISTINCT metadata_collection_keys.metadata_address) AS market_cap
+            FROM listings
+            INNER JOIN metadatas on (listings.metadata = metadatas.address)
+            INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
+            INNER JOIN auction_houses on (listings.auction_house = auction_houses.address)
+            WHERE
+                ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
+                AND auction_houses.treasury_mint = 'So11111111111111111111111111111111111111112'
+                AND listings.created_at >= $2
+                AND listings.created_at <= $3
```
kespinola, line 120:
```suggestion
                AND listings.created_at <= $3
                AND listing.canceled_at IS NULL
                AND listing.purchase_id IS NULL
```

sorry for not getting all my CR in at once but just ran into this. Need to filter out canceled_at and purchase_id to get active listings.

kespinola (changes requested):

crates/core/src/db/queries/collections.rs
austinmilt: @kespinola  Correct me if I'm wrong, but if we only filter to active listings, then `COUNT( DISTINCT metadata_collection_keys.metadata_address)` will omit those NFTs which are not actively listed, which will reduce the total "supply" of NFTs in the collection, which is will result in a calculation different from what we agreed was "market cap".

I guess I can wrap in another `WITH` to get the number of NFTs in a collection without regard to listing status

austinmilt: Without filtering to active listings
image

versus with filtering to active listings
image

austinmilt: First pass at this completely bogs down the query, taking ~21 seconds to complete because of computing collection sizes over the whole table.

```
WITH nft_count_table (collection, nft_count) as (
	SELECT metadata_collection_keys.collection_address AS collection, COUNT( DISTINCT metadata_collection_keys.metadata_address) as nft_count
	FROM metadata_collection_keys
	GROUP BY metadata_collection_keys.collection_address
),

market_cap_table (collection, market_cap) as (
	SELECT metadata_collection_keys.collection_address AS collection, MIN(listings.price) * MAX (nft_count_table.nft_count) AS market_cap
	FROM listings
	INNER JOIN metadatas on (listings.metadata = metadatas.address)
	INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
	INNER JOIN auction_houses on (listings.auction_house = auction_houses.address)
	INNER JOIN nft_count_table on (collection = nft_count_table.collection)
	WHERE
		auction_houses.treasury_mint = 'So11111111111111111111111111111111111111112'
		AND listings.purchase_id IS NULL
		AND listings.canceled_at IS NULL
	GROUP BY metadata_collection_keys.collection_address
	ORDER BY market_cap DESC
	LIMIT 10
	OFFSET 0
) SELECT
	metadatas.name,
	metadatas.seller_fee_basis_points,
	metadatas.update_authority_address,
	metadatas.mint_address,
	metadatas.primary_sale_happened,
	metadatas.uri,
	metadatas.slot,
	metadata_jsons.description,
	metadata_jsons.image,
	metadata_jsons.category,
	metadata_jsons.model
FROM metadata_jsons, market_cap_table, metadatas
WHERE
	market_cap_table.collection = metadatas.mint_address
	AND metadatas.address = metadata_jsons.metadata_address
ORDER BY market_cap DESC
```

kespinola: With feeds there was a big performance increase by splitting the order by from the limit using another with statement. Can you test dropping the limit?

austinmilt: Tried this with no benefit. Will keep poking around

austinmilt: This gets it down to about 5 seconds:

```
WITH market_cap_table (collection, market_cap, nft_count, floor_price) AS (
	SELECT components.collection, components.floor_price * components.nft_count AS market_cap, components.nft_count, components.floor_price
	FROM (
		SELECT floor_prices.collection, MIN(floor_prices.floor_price) AS floor_price, COUNT( DISTINCT metadata_collection_keys.metadata_address) as nft_count
		FROM (
			SELECT metadata_collection_keys.collection_address AS collection, MIN(listings.price) AS floor_price
			FROM listings
			INNER JOIN metadatas on (listings.metadata = metadatas.address)
			INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
			INNER JOIN auction_houses on (listings.auction_house = auction_houses.address)
			WHERE
				auction_houses.treasury_mint = 'So11111111111111111111111111111111111111112'
				AND listings.purchase_id IS NULL
				AND listings.canceled_at IS NULL
			GROUP BY metadata_collection_keys.collection_address
		) floor_prices
		INNER JOIN metadata_collection_keys on (metadata_collection_keys.collection_address = floor_prices.collection)
		GROUP BY floor_prices.collection
	) components
	ORDER BY market_cap DESC
	LIMIT 10
	OFFSET 0
)

SELECT
	market_cap_table.floor_price,
	market_cap_table.nft_count,
	market_cap_table.market_cap,
	metadatas.name,
	metadatas.seller_fee_basis_points,
	metadatas.update_authority_address,
	metadatas.mint_address,
	metadatas.primary_sale_happened,
	metadatas.uri,
	metadatas.slot,
	metadata_jsons.description,
	metadata_jsons.image,
	metadata_jsons.category,
	metadata_jsons.model
FROM metadatas
INNER JOIN metadata_jsons ON (metadata_jsons.metadata_address = metadatas.address)
INNER JOIN market_cap_table ON (market_cap_table.collection = metadatas.mint_address)
ORDER BY market_cap DESC
```

austinmilt: with a time window this gets quicker, but still several seconds

austinmilt:
> Looks good. Can we propagate the new `pre_query_search_limit` parameter to the other Meilisearch queries in `query_root.rs`?

@ray-kast The `pre_query_search_limit` should only be used when passing the results of Meilisearch to an SQL query. That's because this limit is a buffer to help increase the odds of getting consistent results on the SQL query after leveraging Meilisearch. Otherwise, the limit and offset provided in the API should be passed direct to Meilisearch. Currently the other search endpoints have limits and offsets, so no change needs to be made there.

imabdulbasit (approved):

PR #547, opened 2022-06-23
Change latest events api to use include events

Bhargavamacha:


kristianeboe (approved):
I'll probably update some of the variable names / comments back on the dev branch, but good work!

PR #555, opened 2022-06-29
#534: add featured collections (by market cap) query

austinmilt:
This PR adds a query to get collections ordered by market cap (floor price * number of active listings).

# Changes
- added SQL query to get collections ordered by market cap (collections.rs)
- added root Graph QL query for collections by market cap (query_root.rs)

### DIFF
--- a/crates/core/src/db/queries/collections.rs
+++ b/crates/core/src/db/queries/collections.rs
@@ -110,7 +110,7 @@ fn make_by_market_cap_query_string(order_direction: OrderDirection) -> String {
             FROM listing_receipts
             INNER JOIN metadatas on (listing_receipts.metadata = metadatas.address)
             INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
-            WHERE 
+            WHERE
                 ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
                 AND listing_receipts.purchase_receipt IS NULL
                 AND listing_receipts.canceled_at IS NULL
--- a/crates/core/src/db/queries/collections.rs
+++ b/crates/core/src/db/queries/collections.rs
@@ -67,3 +67,58 @@ fn make_by_volume_query_string(order_direction: OrderDirection) -> String {
         order_direction = order_direction
     )
 }
+
+/// Query collections ordered by market cap
+///
+/// # Errors
+/// returns an error when the underlying queries throw an error
+pub fn by_market_cap(
+    conn: &Connection,
+    addresses: impl ToSql<Nullable<Array<Text>>, Pg>,
+    order_direction: OrderDirection,
+    limit: impl ToSql<Integer, Pg>,
+    offset: impl ToSql<Integer, Pg>,
+) -> Result<Vec<Nft>> {
+    diesel::sql_query(make_by_market_cap_query_string(order_direction))
+        .bind(addresses)
+        .bind(limit)
+        .bind(offset)
+        .load(conn)
+        .context("Failed to load collections by volume")
+}
+
+fn make_by_market_cap_query_string(order_direction: OrderDirection) -> String {
+    format!(
+        r"
+    SELECT
+        metadatas.address,
+        metadatas.name,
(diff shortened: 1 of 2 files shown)


crates/core/src/db/queries/collections.rs
```diff
@@ -67,3 +67,61 @@ fn make_by_volume_query_string(order_direction: OrderDirection) -> String {
...
+        metadatas.uri,
+        metadatas.slot,
+        metadata_jsons.description,
+        metadata_jsons.image,
+        metadata_jsons.category,
+        metadata_jsons.model
+    FROM metadata_jsons
+    INNER JOIN metadatas ON (metadatas.address = metadata_jsons.metadata_address)
+    INNER JOIN (
+        SELECT metadata_collection_keys.collection_address AS collection, MIN(listing_receipts.price) * COUNT(listing_receipts.price) AS market_cap
```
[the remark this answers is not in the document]
austinmilt: I'll let @kespinola comment on what market cap should be, but I was told by @bellebellesol to use floor price * supply.

imabdulbasit (approved):

crates/core/src/db/queries/collections.rs
```diff
@@ -67,3 +67,61 @@ fn make_by_volume_query_string(order_direction: OrderDirection) -> String {
...
+            INNER JOIN metadatas on (listings.metadata = metadatas.address)
+            INNER JOIN metadata_collection_keys on (metadatas.address = metadata_collection_keys.metadata_address)
+            WHERE
+                ($1 IS NULL OR metadata_collection_keys.collection_address = ANY($1))
+                AND listings.purchase_id IS NULL
+                AND listings.canceled_at IS NULL
+            GROUP BY metadata_collection_keys.collection_address
+            ORDER BY market_cap {order_direction}
+            LIMIT $2
+            OFFSET $3
```
kespinola, line 120: Same question as before on doing the price market_cap in outer query then joining in the results table int the one returning nfts.

kespinola (changes requested):

crates/core/src/db/queries/collections.rs
kespinola: Screen Shot 2022-06-30 at 2 48 24 PM

There is a long seq scan on `listings.purchase_id IS NULL AND listings.canceled_at IS NULL` that is a pretty common lookup path so good candidate for a joint index.

austinmilt: > Screen Shot 2022-06-30 at 2 48 24 PM
> 
> There is a long seq scan on `listings.purchase_id IS NULL AND listings.canceled_at IS NULL` that is a pretty common lookup path so good candidate for a joint index.



> Screen Shot 2022-06-30 at 2 48 24 PM
> 
> There is a long seq scan on `listings.purchase_id IS NULL AND listings.canceled_at IS NULL` that is a pretty common lookup path so good candidate for a joint index.

This went a little over my SQL-noob head. Are you saying we should create another table/index for active listings, or that I should be doing something additional in my query?

austinmilt: Based on feedback of the market cap calculation, I am omitting these conditions, so not creating the index for this PR.

crates/core/src/db/queries/collections.rs
```diff
@@ -67,3 +67,61 @@ fn make_by_volume_query_string(order_direction: OrderDirection) -> String {
...
+        metadatas.uri,
+        metadatas.slot,
+        metadata_jsons.description,
+        metadata_jsons.image,
+        metadata_jsons.category,
+        metadata_jsons.model
+    FROM metadata_jsons
+    INNER JOIN metadatas ON (metadatas.address = metadata_jsons.metadata_address)
+    INNER JOIN (
+        SELECT metadata_collection_keys.collection_address AS collection, MIN(listings.price) * COUNT(listings.price) AS market_cap
```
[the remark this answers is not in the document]
austinmilt: we are going with floor X number of NFTs in the collection

kespinola (approved):

PR #584, opened 2022-07-18
basic tests

austinmilt:
This PR adds initial setup for running tests against DB queries. Right now it's primarily testing the collections queries. This will ensure that if fields required for deserialing `Nft` change, the build will fail.

### DIFF
--- a/crates/core/src/db/mod.rs
+++ b/crates/core/src/db/mod.rs
@@ -165,3 +165,378 @@ pub fn connect(args: ConnectArgs, mode: ConnectMode) -> Result<(Pool, Connection
 
     Ok((pool, ty))
 }
+
+#[cfg(test)]
+pub mod test {
+    use diesel::{insert_into, prelude::*};
+    use uuid::Uuid;
+
+    use super::{
+        connect, models,
+        schema::{
+            auction_houses, listings, metadata_collection_keys, metadata_jsons, metadatas,
+            purchases,
+        },
+        ConnectArgs,
+    };
+    use crate::prelude::*;
+
+    fn initialize() -> super::Pool {
+        let conn_args = ConnectArgs {
+            database_read_url: None,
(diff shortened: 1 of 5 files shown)


mpwsh (approved):

crates/core/test_migrations/2022-08-04-200854_seed_nfts/down.sql
```diff
@@ -0,0 +1,27 @@
+delete from metadatas where address = any(array [
```
kespinola, line 1: Normally unit tests involving databases are done in transaction that leave the db empty after each unit test but since we are only doing read operations seeding with a shared set of data works for our use case.

crates/core/src/db/mod.rs
```diff
@@ -187,25 +210,64 @@ pub fn connect(args: ConnectArgs, mode: ConnectMode) -> Result<ConnectResult> {
...
+#[cfg(test)]
+pub mod test {
+    embed_migrations!("test_migrations");
 
-        embedded_migrations::run_with_output(
-            &pool.get().context("Failed to connect to the database")?,
-            &mut out,
+    fn initialize() -> super::ConnectResult {
+        dotenv::from_filename(".env.dev").expect("Failed to load .env.dev");
+        dotenv::from_filename(".env").expect("Failed to load .env");
```
austinmilt, line 231: should this also read from `.env.local`?

crates/core/test_migrations/2022-08-04-200854_seed_nfts/down.sql
austinmilt: Seconding this. I like the migration setup. It would be awesome to be able to define a migration per test in such a way that we could run `up` and `down` for a specific migration for each test, just to keep the db minimally populated with test-specific data.

This is a great start, though!

PR #606, opened 2022-07-20
add offer activity

dandlezzz:
YOU WANT AN UNCHECKED QUERY? YOU GOT IT.

Need offers in order to support the activity features outlined [here](https://github.com/holaplex/holaplex/issues/607)


mpwsh (approved):

kespinola (approved):
got to run cargo fmt.

kespinola (commented):
Note to check marketplaces to make sure this api change doesn't break its activity. We don't have to account for it now but shouldn't let it break that activity stream.

PR #612, opened 2022-07-21
WIP: collection nft count and floor price

austinmilt:
TBD

### DIFF
--- a/crates/graphql/src/schema/objects/nft.rs
+++ b/crates/graphql/src/schema/objects/nft.rs
@@ -295,7 +295,7 @@ impl NftActivity {
 
 #[graphql_interface(for = [Nft])]
 #[async_trait]
-trait NftTraits {
+pub trait NftTraits {
     
     fn address(&self) -> &str;
     fn name(&self) -> &str;
@@ -577,7 +577,7 @@ impl NftTraits for Nft {
 
 
 #[graphql_interface(for = CollectionNft)]
-trait CollectionNftTraits {
+pub trait CollectionNftTraits {
     fn nft_count(&self, context: &AppContext) -> FieldResult<i32>;
 
     //TODO floor price
--- a/crates/graphql/src/schema/objects/nft.rs
+++ b/crates/graphql/src/schema/objects/nft.rs
@@ -1,8 +1,9 @@
+use async_trait::async_trait;
 use indexer_core::{
     assets::{proxy_url, AssetIdentifier, ImageSize},
     db::{
         queries,
-        tables::{bid_receipts, listing_receipts, metadata_jsons},
+        tables::{bid_receipts, listing_receipts, metadata_jsons, metadata_collection_keys},
     },
     util::unix_timestamp,
     uuid::Uuid,
@@ -14,6 +15,7 @@ use objects::{
 use reqwest::Url;
 use scalars::{PublicKey, U64};
 use serde_json::Value;
+use juniper::graphql_interface;
 
 use super::prelude::*;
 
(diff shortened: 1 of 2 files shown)


crates/graphql/src/schema/objects/nft.rs
```diff
@@ -521,9 +575,30 @@ If no value is provided, it will return XSmall")))]
...
+    fn nft_count(&self, context: &AppContext) -> FieldResult<i32> {
+        let conn = context.shared.db.get()?;
+
+        let count: i64 = metadata_collection_keys::table.filter(metadata_collection_keys::collection_address.eq(&self.0.mint_address))
+            .count()
+            .get_result(&conn)
+            .context("failed to load NFT count")?;
+
+        Ok(count.try_into()?)
+    }
```
kespinola, line 598: This needs to be in a data loader or we'll get n+1 queries.

kespinola (changes requested):

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -290,8 +292,55 @@ impl NftActivity {
     }
 }
 
-#[derive(Debug, Clone)]
+
+#[graphql_interface(for = [Nft])]
+#[async_trait]
+pub trait NftTraits {
```
kespinola, line 298: nice.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -522,69 +574,290 @@ If no value is provided, it will return XSmall")))]
...
+        let conn = context.shared.db.get()?;
 
-    fn try_from(value: models::Nft) -> Result<Self, Self::Error> {
-        value.try_into().map(Self)
+        let count = collection_stats::table
+            .filter(collection_stats::collection_address.eq(&self.0.mint_address))
+            .select(collection_stats::nft_count)
+            .get_result::<i64>(&conn)
+            .context("failed to load NFT count for this collection")?;
+
```
kespinola, line 589: Should do this in a dataloader to avoid n+1 when collections queried in a list.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -522,69 +574,290 @@ If no value is provided, it will return XSmall")))]
...
-    fn mark() {
-        <Nft as juniper::marker::IsOutputType<S>>::mark();
+    fn floor_price(&self, context: &AppContext) -> FieldResult<Option<i32>> {
+        let conn = context.shared.db.get()?;
+
+        let count = collection_stats::table
+            .filter(collection_stats::collection_address.eq(&self.0.mint_address))
+            .select(collection_stats::floor_price)
+            .get_result::<Option<i64>>(&conn)
+            .context("failed to load floor price for this collection")?;
```
kespinola, line 600: another spot for data_loader.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -290,9 +292,58 @@ impl NftActivity {
     }
 }
 
-#[derive(Debug, Clone)]
+#[graphql_interface(for = [BaseNft, CollectionNft])]
+#[async_trait]
+pub trait Nft {
```
kespinola, line 297: NftExt of AsNft or BaseNft

I'd like to keep Nft holding its name its less confusing to developers. I was researching some trait naming conventions thinking NftExt is the most appropriate here.
https://internals.rust-lang.org/t/naming-convention-for-traits/1796/9
https://github.com/rust-lang/api-guidelines/discussions/28#discussioncomment-233418

kespinola (changes requested):
This is looking really good. great job. blocking to implement dataloader for nft_counts and floor. I'd also like to keep an Nft called nft.

crates/graphql/src/schema/objects/nft.rs
austinmilt: I've been trying to figure out how to implement this in a way that does not break Holaplex.com and marketplaces. So far this implementation is the closest, yet still falls short because of the issue I posted in slack. Renaming the trait to NftExt will also lead to breaks.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -16,6 +18,10 @@ use scalars::{PublicKey, U64};
 use serde_json::Value;
 
 use super::prelude::*;
+use crate::schema::{
+    dataloaders::collection::{CollectionFloorPrice, CollectionNftCount},
+    scalars::I64,
+};
```
kespinola, line 24: should be able to reference without the use create::schema

Just `use scalars::I64` and `use dataloaders::collection::{...}`

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -352,67 +409,68 @@ impl TryFrom<models::Nft> for Nft {
...
+    fn _model(&self) -> Option<String> {
+        self.model.as_ref().map(ToString::to_string)
+    }
+
+    fn _address(&self) -> &str {
         &self.address
     }
 
-    pub fn name(&self) -> &str {
+    fn _name(&self) -> &str {
```
kespinola, line 429: Why are all of these _ now?

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -521,70 +571,514 @@ If no value is provided, it will return XSmall")))]
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct CollectionNft(Nft);
+#[graphql_object(impl = NftExtValue, Context = AppContext)]
+impl Nft {
```
kespinola, line 575: Is the double imp needed?

kespinola (commented):
Is duplication not cleaner than the use of a trait? Like to discuss this today during tech sync.

crates/graphql/src/schema/objects/nft.rs
austinmilt: See code comments and PR description. In short: having two (nearly) identical `impl` blocks for `Nft` and `CollectionNft` was required so Apollo could pick up the available fields when generating client code

crates/graphql/src/schema/objects/nft.rs
austinmilt: Two issues:

1. I ended up having to do 4 `impl` of the same methods and wanted a single "source" implementation to refer to. Compounds with
2. Rust wasnt able to determine that the `impl Nft` block without underscore method names was the one I was referring to when doing e.g. `self.name()` or `self.0.name()` in the `impl NftExt` blocks (since "self" in those blocks refers to `NftExt` rather than the struct for which the trait is being implemented). This led to infinite recursion.

This was my workaround :/

kespinola (approved):

PR #617, opened 2022-07-22
add marketplace_address field 

imabdulbasit:
MagicEden auctionhouse data is not indexed in auction_houses table so marketplace_program_address field is added to expose the program address of marketplaces.

### DIFF
--- a/crates/graphql/src/schema/objects/ah_offer.rs
+++ b/crates/graphql/src/schema/objects/ah_offer.rs
@@ -11,6 +11,7 @@ pub struct Offer {
     pub buyer: PublicKey<Wallet>,
     pub metadata: PublicKey<Nft>,
     pub auction_house: PublicKey<AuctionHouse>,
+    pub marketplace_program_address: String,
     pub price: U64,
     pub purchase_id: Option<Uuid>,
     pub trade_state_bump: i32,
@@ -67,6 +68,10 @@ impl Offer {
         self.token_size
     }
 
+    fn marketplace_program_address(&self) -> &str {
+        &self.marketplace_program_address
+    }
+
     pub async fn nft(&self, ctx: &AppContext) -> FieldResult<Option<Nft>> {
         ctx.nft_loader
             .load(self.metadata.clone())
@@ -90,6 +95,7 @@ impl<'a> TryFrom<models::Offer<'a>> for Offer {
             id,
             trade_state,
             auction_house,
+            marketplace_program,
             buyer,
             metadata,
             token_account,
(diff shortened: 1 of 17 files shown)


crates/graphql/src/schema/objects/nft.rs
```diff
@@ -235,7 +236,8 @@ impl TryFrom<models::NftActivity> for NftActivity {
         Ok(Self {
             id,
             metadata: metadata.into(),
-            auction_house: auction_house.into(),
+            auction_house: auction_house.clone().into(),
+            marketplace_address: auction_house,
```
kespinola, line 240: hmm this doesn't sit right. we should update the NftActivity query to return martplace_address.

crates/graphql/src/schema/objects/ah_offer.rs
```diff
@@ -39,6 +40,10 @@ impl Offer {
         &self.metadata
     }
 
+    fn marketplace_address(&self) -> &str {
```
kespinola, line 43: marketplace_program_address is a more descriptive naming for the field.

kespinola (changes requested):
Also be good to add a migration that moves the existing ME listing, purchases, and offers auction_house address to marketplace_address 

When we index the accounts should write this fields in so expecting a new column on the tables.

imabdulbasit:
> Also be good to add a migration that moves the existing ME listing, purchases, and offers auction_house address to marketplace_address
> 
> When we index the accounts should write this fields in so expecting a new column on the tables.

I would say we rename the auction_house column to marketplace_program. The marketplace_program address will then be loaded into the ah dataloader we already have

crates/indexer/src/geyser/programs/magic_eden_haus.rs
```diff
@@ -46,7 +49,8 @@ async fn process_execute_sale(
             id: None,
             buyer: Owned(accts[0].clone()),
             seller: Owned(accts[1].clone()),
-            auction_house: Owned(accts[9].clone()),
+            auction_house: None,
```
kespinola, line 52: Since ME is a fork of auction house I really do think one of the accounts is an auction house. It be a massive refactor to remove it and rather fruitless.

kespinola (approved):

kespinola (approved):

milorue (approved):

milorue:
Does this also fix the open sea auction house data?

imabdulbasit:
> Does this also fix the open sea auction house data?

https://github.com/holaplex/indexer/pull/633
this PR will fix that.

PR #620, opened 2022-07-22
add allow_unverified filter on nfts query

imabdulbasit:


crates/core/src/db/queries/metadatas.rs
```diff
@@ -291,8 +294,11 @@ pub fn list(
...
                     .equals(MetadataCreators::Table, MetadataCreators::MetadataAddress),
             )
-            .and_where(Expr::col(MetadataCreators::CreatorAddress).is_in(creators))
-            .and_where(Expr::col(MetadataCreators::Verified).eq(true));
+            .and_where(Expr::col(MetadataCreators::CreatorAddress).is_in(creators));
+    }
+
+    if !matches!(allow_unverified, Some(true)) {
+        query.and_where(Expr::col(MetadataCreators::Verified).eq(true));
     }
```
kespinola, line 302:
```suggestion
    if allow_unverified != Some(true) {
        query.and_where(Expr::col(MetadataCreators::Verified).eq(true));
    }
```

Does this not work. reads better.

kespinola (approved):

PR #631, opened 2022-07-27
Add cancel_listing event in wallet activities

alchemistgo87:


imabdulbasit (approved):

imabdulbasit (commented):
We might need to expose marketplace_program for wallet_activity

PR #642, opened 2022-08-01
#630: only do collection stats migrations if needed

austinmilt:
This adds a condition to only do the `collection_stats` migrations if it hasnt been run before (by checking that there are >0 rows in the table). This way we can run the backfill manually before deploying prod to prevent this potentially long-running migration


kespinola (dismissed):

imabdulbasit (approved):

dandlezzz (approved):
approved.

Issue #576, opened 2022-07-13
Expose NFT Symbol Through The API


kespinola:
### Goal
Client can query symbol field on the NFT.

### Requirement
Expose the symbol field on the NFT graphql object.

PR #643, opened 2022-08-02
expose nft symbol through API

wbarksdale:
Wanted to try a small indexer ticket and found this one. I basically just copied the pattern from https://github.com/holaplex/indexer/issues/576

Testing: 
ran `./scripts/pre-push.sh` and it passed

Closes: #576


wbarksdale:
looks like this was already done here: https://github.com/holaplex/indexer/pull/610

PR #651, opened 2022-08-04
implements query for a candy machine by address

wbarksdale:
Implements a query for candy machines by the candy machine's address. Joins the data from `candy_machines` and `candy_machines_datas` table into a single GraphQL object.

Testing: 
pre-push checks succeeded.

### DIFF
--- a/crates/core/src/db/models.rs
+++ b/crates/core/src/db/models.rs
@@ -830,6 +830,47 @@ pub struct GraphConnection<'a> {
     pub write_version: i64,
 }
 
+/// A join of tables `candy_machines` and `candy_machine_datas`
+#[derive(Debug, Clone, Queryable)]
+#[diesel(treat_none_as_null = true)]
+pub struct CandyMachineJoined {
+    // Table 'candy_machines'
+    /// CandyMachine account address
+    pub address: String,
+    /// CandyMachine 'Authority' address
+    pub authority: String,
+    /// CandyMachine 'Wallet' address
+    pub wallet: String,
+    /// Token mint address
+    pub token_mint: Option<String>,
+    /// Items redeemed
+    pub items_redeemed: i64,
+
+    // Table 'candy_machine_data'
+    /// CandyMachine account address
+    pub candy_machine_address: String,
+    /// Uuid
+    pub uuid: String,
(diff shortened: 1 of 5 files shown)


crates/core/src/db/models.rs
```diff
@@ -830,6 +830,47 @@ pub struct GraphConnection<'a> {
     pub write_version: i64,
 }
 
+/// A join of tables `candy_machines` and `candy_machine_datas`
+#[derive(Debug, Clone, Queryable)]
+#[diesel(treat_none_as_null = true)]
+pub struct CandyMachineJoined {
```
wbarksdale, line 836: Does this seem correct? @echohtp

crates/graphql/src/schema/objects/candymachine.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub is_mutable: bool,
+    pub retain_authority: bool,
+    pub go_live_date: Option<U64>,
+    pub items_available: U64,
+}
+
+impl<'a, 'b> TryFrom<(models::CandyMachine<'a>, models::CandyMachineData<'b>)> for CandyMachine {
+    type Error = std::num::TryFromIntError;
+
+    fn try_from(
```
wbarksdale, line 50: @ray-kast I removed the `CandyMachineJoined` models object, and used a tuple of tuples in the `select` statement instead.

crates/graphql/src/schema/scalars/mod.rs
```diff
@@ -10,6 +10,9 @@ pub(self) mod prelude {
 pub mod markers {
     pub struct StoreConfig;
     pub struct TokenMint;
+    /// Used to denote a public key field to graphql while not
+    /// indicating a relationship to any other specific object
+    pub struct Unspecified;
```
wbarksdale, line 15: @ray-kast I can delete this if its dumb / not useful

wbarksdale: I was thinking that `Wallet` sort of implies a "user" since it is linked with a twitter account, when that isn't necessarily the case here, but I deleted this `Unspecified` marker and replaced with `Wallet`

echohtp (approved):

PR #672, opened 2022-08-11
fix handling of candy machine config lines

wbarksdale:
Candy machine config line parsing previously assumed that config line's were stored in their own PDA, however these config lines are actually stored as part of the candy machine account itself, but are not deserialized by mpl's candy machine data structure.

This implements a function to parse those config lines. The config lines are indexed based on a compound key `(candy_machine_address, idx)` where idx is the the index of the config line within the candy machine account data. 

I also took a stab at adding a test for this, its not a super useful test, but wanted to give it a shot. 
In order to make the test work, I needed to add a feature `test-internal` to expose the internals of the indexer to the test, this resulted in some changes to `pre-push.sh` and `Cargo.toml`. I'm happy to delete this test if its not wanted, but thought i'd take a stab at it.

### DIFF
--- a/crates/indexer/src/geyser/programs/candy_machine.rs
+++ b/crates/indexer/src/geyser/programs/candy_machine.rs
@@ -35,12 +35,12 @@ pub fn parse_cm_config_lines(
     // This could occur if this function is called on a candy machine that uses hiddensettings instead of
     // config lines
     let bytes_needed_for_taken_bitmask = items_available / 8 + (items_available % 8).min(1);
-    let expected_taken_len = taken_bitmask_start + bytes_needed_for_taken_bitmask;
+    let total_expected_data_len = taken_bitmask_start + bytes_needed_for_taken_bitmask;
 
-    if expected_taken_len >= data.len() {
+    if total_expected_data_len >= data.len() {
         bail!(
             "Config line bytes would overflow available data ({} vs {})",
-            expected_taken_len,
+            total_expected_data_len,
             data.len()
         );
     }
@@ -87,7 +87,7 @@ pub async fn process_cm(client: &Client, update: AccountUpdate) -> Result<()> {
     let items_available = usize::try_from(candy_machine.data.items_available)
         .context("Failed to convert available item count")?;
 
-    let lines = if candy_machine.data.hidden_settings.is_some() {
+    let lines = if candy_machine.data.hidden_settings.is_none() {
         Some(
             parse_cm_config_lines(&update.data, items_available)
                 .context("Failed to parse candy machine lines")?,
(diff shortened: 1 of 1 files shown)


crates/core/migrations/2022-08-12-183150_add_index_and_taken_to_config_lines/down.sql
```diff
@@ -0,0 +1,13 @@
...
+drop column taken;
+
+alter table candy_machine_config_lines
+rename column candy_machine_address to address;
+
+alter table candy_machine_config_lines
+alter table drop constraint candy_machine_address_idx_pkey;
+
+alter table candy_machine_config_lines
+add constraint candy_machine_config_lines_pkey primary key (address);
```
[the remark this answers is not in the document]
wbarksdale: for some reason it seems to want the `rename` to be separate?

wbarksdale:
@ray-kast I think I got everything, I left the data the test relies on in `crates/indexer/tests/data` because this seems to be the `current_dir()` for when the test is run. I also cleaned up the sql a bit, but for some reason it seemed to want the column rename to be in its own `alter table`, i'm not sure why, but let me know if you have any insight or want further changes there.

crates/indexer/src/geyser/programs/candy_machine.rs
```diff
@@ -87,7 +87,7 @@ pub async fn process_cm(client: &Client, update: AccountUpdate) -> Result<()> {
     let items_available = usize::try_from(candy_machine.data.items_available)
         .context("Failed to convert available item count")?;
 
-    let lines = if candy_machine.data.hidden_settings.is_some() {
+    let lines = if candy_machine.data.hidden_settings.is_none() {
```
wbarksdale, line 90: @imabdulbasit this was the issue, just needed to invert this logic

crates/indexer/src/geyser/programs/candy_machine.rs
```diff
@@ -35,12 +35,12 @@ pub fn parse_cm_config_lines(
     // This could occur if this function is called on a candy machine that uses hiddensettings instead of
     // config lines
     let bytes_needed_for_taken_bitmask = items_available / 8 + (items_available % 8).min(1);
-    let expected_taken_len = taken_bitmask_start + bytes_needed_for_taken_bitmask;
+    let total_expected_data_len = taken_bitmask_start + bytes_needed_for_taken_bitmask;
```
wbarksdale, line 38: renamed this for clarity since "taken" in this context refers to the length of the "taken" bitmask

crates/indexer/src/geyser/programs/candy_machine.rs
```diff
@@ -138,15 +138,18 @@ mod tests {
             "candy_machines/AoHidoffmkL4xURViNgbA4YyeDw82FAYUZfomL3X5BoU.dmp",
             "candy_machines/piA76RvvmCt7UWEmJSBVA6xMoXqwvEAELwJoqeHK6i3.dmp",
             "candy_machines/CiBuYi3W3aVQbMWcjvfKBpwjHS6fViuuxQdSUUqkjkn4.dmp",
-            "candy_machines/ACDPaQ3uGy33KsBKiUH4azDX4q7Nxk3QwW3trALEdFmB.dmp", /* this one has hidden settings */
+            // This candy machine has hidden settings, but for some reason has allocated space
+            // for all the config lines that would be needed, however they are all empty.
+            // This results in returning a results array with zero as none of the config
+            // lines are marked as available
```
wbarksdale, line 144: just to add the the annoyingness of candy machines, some (but not all) of them seem to have erroneously allocated space for config lines, but still have hidden settings.

wbarksdale (commented):
Tested this locally and it is ingesting config lines

imabdulbasit (approved):

PR #673, opened 2022-08-15
use compound key for candy machine creators

wbarksdale:
Candy machines can have multiple, so using the candy_machine_address alone as the primary key doesn't work. This changes the table to use a compound primary key of `(candy_machine_address, creator_address)`

### DIFF
--- a/crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/down.sql
+++ b/crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/down.sql
@@ -1,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+alter table candy_machine_creators
+drop constraint candy_machine_creators_compound_pkey,
+add constraint candy_machine_creators_pkey primary key (candy_machine_creators_compound_pkey);
--- a/crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/up.sql
+++ b/crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/up.sql
@@ -1,0 +1,4 @@
+-- Your SQL goes here
+alter table candy_machine_creators
+drop constraint candy_machine_creators_pkey,
+add constraint candy_machine_creators_compound_pkey primary key (candy_machine_address, creator_address);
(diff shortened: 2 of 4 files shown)


imabdulbasit (dismissed):

crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Your SQL goes here
+alter table candy_machine_creators
+drop constraint candy_machine_creators_pkey,
+add constraint candy_machine_creators_compound_pkey primary key (candy_machine_address, creator_address);
```
wbarksdale, line 4: hmm, do we care that this can fail when migrating backwards?

imabdulbasit: you should just add `select false` in down.sql because the on_conflict code need to be changed as well if the migration is reverted

crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/down.sql
```diff
@@ -0,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+alter table candy_machine_creators
+drop constraint candy_machine_creators_compound_pkey,
+add constraint candy_machine_creators_pkey primary key (candy_machine_creators_compound_pkey);
```
imabdulbasit, line 4: Why is `candy_machine_creators_compound_pkey` added as primary key? This should be candy_machine_address if you want to revert it back

wbarksdale: yeah, that was my intention, oops!

crates/core/migrations/2022-08-15-172917_add_compound_key_to_candymachine_creators/down.sql
```diff
@@ -0,0 +1,7 @@
+-- drop duplicates from candy_machine creators
+delete from candy_machine_creators a using candy_machine_creators b
+where a.candy_machine_address=b.candy_machine_address and a.creator_address > b.creator_address;
```
wbarksdale, line 3: @ray-kast how is this?

imabdulbasit: Yes

wbarksdale:
@ray-kast @imabdulbasit this needs and approval to merge, I think its looking good yeah?

imabdulbasit (approved):

PR #674, opened 2022-08-15
mark missing nfts as burned

### LINKED ISSUE #646: Handle previously burned NFTs
Create a task that uses json-rpc to remove metadata records (and dependent records) where the token has been burned.

dandlezzz:
closes #646


burn_fix/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[package]
+name = "holaplex-rpc-client"
+version = "0.0.0"
+edition = "2021"
+
+
+[dependencies]
+solana-client = "1.7.11"
+solana-program = "1.7.11"
+anyhow = "1.0.45"
```
imabdulbasit, line 13: anyhow is already imported using the core crate

burn_fix/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+            let addressess = metadatas::table.filter(metadatas::burned_at.is_null())
+                .offset(i)
+                .limit(limit)
+                .order_by(metadatas::slot.asc())
+                .select(metadatas::address)
+                .load(&conn)
+                .context("could not load metadatas")?;
+
+            addressess.into_iter().try_for_each(|address: String| {
+                let key: Pubkey = address.parse().unwrap();
```
imabdulbasit, line 47: unwrap() should be removed to avoid panic. You can use context() from anyhow

imabdulbasit: We need to keep track of NFTs that have been processed so that if we start again we know where to start

imabdulbasit (changes requested):

burn_fix/src/main.rs
imabdulbasit: What do you think about updated_at optional timestamp to be added into metadatas. We can then process metadatas `where burned_at is null AND updated_at is null`

imabdulbasit (approved):

PR #681, opened 2022-08-17
Fetch collected collections with the wallet api

alchemistgo87:
### Changes
- Query collectedCollections for a wallet including number of nfts owned from the collection by the wallet and the estimated value of the holder's position in the collection based on its floor price.
Screen Shot 2022-08-23 at 2 47 02 PM


kespinola (changes requested):

kespinola (changes requested):
Lets just get the collection query in a db query then good to go.

The other item is to see if the query can be done with diesel. That is ideal as the type checking ensures we don't break queries with future changes. That has been an issue with our by hand sql.

kespinola (approved):

PR #690, opened 2022-08-22
dao-marketplace/#45: collection stats

austinmilt:
This PR adds three new fields to the `Collection` object to support collection display in the marketplace.

# Changes
- added query and field for total volume of sales, with time range args (`collections.rs`, `nft.rs`)
- added query and field for count of current holders of collection NFTs, which throws an error if the collection isnt found (`collections.rs`, `nft.rs`)
- added query and field for count of active listings of collection NFTs, which throws an error if the collection isnt found (`collections.rs`, `nft.rs`)

### DIFF
--- a/crates/core/src/db/queries/collections.rs
+++ b/crates/core/src/db/queries/collections.rs
@@ -1,10 +1,10 @@
 //! Query utilities for collections.
 
-use anyhow::Context;
+use anyhow::{bail, Context};
 use chrono::{DateTime, Utc};
 use diesel::{
     pg::Pg,
-    sql_types::{Array, Integer, Nullable, Text, Timestamp},
+    sql_types::{Array, BigInt, Integer, Nullable, Text, Timestamp},
     types::ToSql,
     RunQueryDsl,
 };
@@ -212,7 +212,7 @@ SELECT listings.id as id, metadata, auction_house, price, created_at, marketplac
     OFFSET $4;
 
  -- $1: address::text
- -- $2: evnet_types::text[]
+ -- $2: event_types::text[]
  -- $3: limit::integer
  -- $4: offset::integer";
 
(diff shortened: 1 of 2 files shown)


crates/core/src/db/queries/collections.rs
```diff
@@ -235,3 +235,113 @@ pub fn collection_activities(
...
+WHERE
+    metadata_collection_keys.collection_address = $1
+	AND metadatas.burned_at IS NULL
+	AND metadata_collection_keys.verified = true
+GROUP BY metadata_collection_keys.collection_address;
+
+-- $1: address::text";
+
+/// Computes the total of all sales of all NFTs in the collection over all time, in lamports.
+///
```
kespinola, line 326: Description is the one for volume total. Should be total of unique holdeers of NFTs.

crates/core/src/db/queries/collections.rs
```diff
@@ -235,3 +235,113 @@ pub fn collection_activities(
...
+const VOLUME_TOTAL_QUERY: &str = r"
+SELECT CAST(SUM(purchases.price) AS BIGINT) AS value
+FROM purchases
+INNER JOIN metadatas ON(purchases.metadata = metadatas.address)
+INNER JOIN metadata_collection_keys ON(metadatas.address = metadata_collection_keys.metadata_address)
+INNER JOIN auction_houses ON(purchases.auction_house = auction_houses.address)
+WHERE
+    metadata_collection_keys.collection_address = $1
+    AND purchases.created_at >= $2
+    AND purchases.created_at <= $3
```
kespinola, line 280: Appreciate the flexibility of setting the time frame but to simplify the requirements for the client I recommend we drop the start and end date range and return all purchases since we are using for total volume.

crates/core/src/db/queries/collections.rs
```diff
@@ -235,3 +235,113 @@ pub fn collection_activities(
...
+-- $1: address::text";
+
+/// Computes the total of all sales of all NFTs in the collection over all time, in lamports.
+///
+/// # Errors
+/// This function fails if the underlying SQL query returns an error
+pub fn holder_count(conn: &Connection, address: impl ToSql<Text, Pg>) -> Result<i64> {
+    let query_result = diesel::sql_query(HOLDER_COUNT_QUERY)
+        .bind(address)
+        .load::<BigIntWrapper>(conn)
```
kespinola, line 332: If the first record is grabbed then shouldn't need the BigIntWrapper can just use BigInt and won't need to use `.get(0).ok_ ...`

Look for uses of `.first()` in the queries.

kespinola (changes requested):
Fine for speed writing the queries as sql strings but doing them through diesel dsl helps ensure further updates don't break the query as the type system validates the queries. I think all these queries are supported by diesel.

austinmilt:
> Fine for speed writing the queries as sql strings but doing them through diesel dsl helps ensure further updates don't break the query as the type system validates the queries. I think all these queries are supported by diesel.

Great point! I was thinking I needed `GROUP BY` which diesel doesnt support, but since it's just a single collection, the `GROUP BY` shouldnt be necessary. I'll refactor

crates/core/src/db/queries/collections.rs
austinmilt: aye aye; it was part of the ticket, but I will remove

kespinola (dismissed):
epic!

imabdulbasit (approved):

PR #698, opened 2022-08-25
Refactor Wallet Activities Query

alchemistgo87:
- Add limit, offset, event types as variables to wallet activities query
- Remove cancelled_listing event from wallet activities query

### DIFF
--- a/crates/core/src/db/queries/wallet.rs
+++ b/crates/core/src/db/queries/wallet.rs
@@ -74,16 +74,13 @@ pub fn activities(
     limit: impl ToSql<Integer, Pg>,
     offset: impl ToSql<Integer, Pg>,
 ) -> Result<Vec<WalletActivity>> {
-    let query = diesel::sql_query(ACTIVITES_QUERY)
+    diesel::sql_query(ACTIVITES_QUERY)
         .bind(address)
         .bind(event_types)
         .bind(limit)
         .bind(offset)
         .load(conn)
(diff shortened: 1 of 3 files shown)


kespinola (approved):
nice. way to get the pagination and event_types in 👍

kespinola (changes requested):
@alchemistgo87 what is using this query right now? I'm think what UI will break from this change.

alchemistgo87:
> @alchemistgo87 what is using this query right now? I'm think what UI will break from this change.

I suppose right now this can only break the current profile activity page at holaplex.com

crates/core/src/db/queries/wallet.rs
```diff
@@ -10,62 +15,75 @@ use crate::{
...
+) -> Result<Vec<WalletActivity>> {
+    let query = diesel::sql_query(ACTIVITES_QUERY)
         .bind(address)
+        .bind(event_types)
+        .bind(limit)
+        .bind(offset)
         .load(conn)
-        .context("Failed to load wallet(s) activities")
+        .context("Failed to load wallet(s) activities");
+    println!("Query Result: {:?}", query);
```
kespinola, line 84: This println should be dropped.

kespinola:
https://github.com/holaplex/holaplex/pull/701

Holaplex updated to work with pagination.

kespinola (approved):

PR #701, opened 2022-08-29
Fetch created collections with wallet query

alchemistgo87:
### Goal
Query created collections by a wallet. A created collection is one in which the wallet is the update authority on the collection nft.

### Changes
- collection query for finding created collections based on the update authority
- Switch the collection field on CreatedCollection and CollectedCollection to use nft data loader.


crates/core/src/db/queries/wallet.rs
```diff
@@ -114,3 +114,34 @@ pub fn collected_collections(
...
+    metadata_collection_keys.collection_address as collection,
+	COUNT(metadatas.address) as nfts_created
+    FROM metadatas
+    INNER JOIN metadata_creators ON (metadata_creators.metadata_address = metadatas.address)
+    INNER JOIN metadata_jsons ON (metadata_jsons.metadata_address = metadatas.address)
+    INNER JOIN metadata_collection_keys ON (metadata_collection_keys.metadata_address = metadatas.address)
+	INNER JOIN collection_stats ON (metadata_collection_keys.collection_address = collection_stats.collection_address)
+    INNER JOIN metadatas collection_metadatas ON (collection_metadatas.mint_address = metadata_collection_keys.collection_address)
+	INNER JOIN metadata_jsons collection_metadata_jsons ON (collection_metadata_jsons.metadata_address = collection_metadatas.address)
+    WHERE metadata_creators.creator_address = $1
```
kespinola, line 129: AND metadata_creators.verified

put verified check as well.

kespinola (changes requested):
Looking more at the requirements all that is needed is the nft count and floor price which is available on collection_stats.

This query should join in metadata_collection_keys and pick the collection nft where the current wallet is the updateAuthority on the collection NFT.

This query should just return Collection model.

Screen Shot 2022-08-29 at 4 44 48 PM

crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -167,6 +167,35 @@ impl CollectedCollection {
...
+#[graphql_object(Context = AppContext)]
+impl CreatedCollection {
+    async fn collection(&self, ctx: &AppContext) -> FieldResult<Option<Collection>> {
+        let conn = ctx.shared.db.get()?;
+
+        queries::collections::get(&conn, &self.collection)?
+            .map(TryInto::try_into)
+            .transpose()
+            .map_err(Into::into)
+    }
```
kespinola, line 196: This is an n+1 this should be a data loader

kespinola (changes requested):
need data loader for the collection there may already be one.

kespinola (approved):

kespinola (approved):
Just looking for feedback on the nested maps I commented on.

PR #704, opened 2022-08-31
ME non-MCC Collections

imabdulbasit:


crates/graphql/src/schema/objects/nft.rs
```diff
@@ -637,27 +648,33 @@ impl Collection {
     #[graphql(
         description = "Total of all sales of all NFTs in the collection over all time, in lamports."
     )]
-    pub async fn volume_total(&self, ctx: &AppContext) -> FieldResult<scalars::U64> {
+    pub async fn volume_total(&self, ctx: &AppContext) -> FieldResult<Option<scalars::U64>> {
```
kespinola, line 651: heads up we are moving to thirty_day_volume and need to handle this as a data loader as it will be returned from a list endpoint.

I was working on that but will hold off.

kespinola (approved):

mpwsh (approved):

PR #710, opened 2022-08-31
update collection_stats functions and triggers

imabdulbasit:


kespinola (changes requested):
This needs to be done in a new migration or it wont apply on db that already have the migration executed.

kespinola (dismissed):
normally need new migration but we have 2 dbs that were updated by hand. Any local dbs will need to be dropped and recreated. 

Anyone booting a fresh db won't have a problem 👍

imabdulbasit:
> normally need new migration but we have 2 dbs that were updated by hand. Any local dbs will need to be dropped and recreated.
> 
> Anyone booting a fresh db won't have a problem +1

or the collection_stats table can truncated and migration can be  run manually

kespinola (approved):

PR #732, opened 2022-09-09
Collection nfts query

alchemistgo87:


crates/core/src/db/queries/metadatas.rs
```diff
@@ -426,6 +429,164 @@ pub fn list(
...
+            Condition::all()
+                .add(Expr::tbl(Listings::Table, Listings::PurchaseId).is_null())
+                .add(Expr::tbl(Listings::Table, Listings::CanceledAt).is_null())
+                .add(
+                    Expr::tbl(Listings::Table, Listings::Expiry)
+                        .is_null()
+                        .or(Expr::tbl(Listings::Table, Listings::Expiry).gt(current_time)),
+                ),
+        )
+        .take();
```
kespinola, line 474: Just need a left join instead of lateral since only 1 marketplace can be selected at a time.

crates/graphql/src/schema/enums.rs
```diff
@@ -17,3 +17,30 @@ impl From<OrderDirection> for db::custom_types::OrderDirection {
...
+        match other {
+            OrderDirection::Desc => Self::Desc,
+            OrderDirection::Asc => Self::Asc,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, juniper::GraphQLEnum)]
+#[graphql(description = "Sorts results by price or listed at")]
+pub enum Sort {
```
kespinola, line 32:
```suggestion
pub enum CollectionNFTSort {
```

crates/core/src/db/queries/metadatas.rs
```diff
@@ -503,79 +509,124 @@ pub fn collection_nfts(conn: &Connection, options: CollectionNftOptions) -> Resu
...
+            ])
+            .columns(vec![
+                (MetadataJsons::Table, MetadataJsons::Description),
+                (MetadataJsons::Table, MetadataJsons::Image),
+                (MetadataJsons::Table, MetadataJsons::AnimationUrl),
+                (MetadataJsons::Table, MetadataJsons::ExternalUrl),
+                (MetadataJsons::Table, MetadataJsons::Category),
+                (MetadataJsons::Table, MetadataJsons::Model),
+            ])
+            .columns(vec![(
```
kespinola, line 533: The columns selected in the me nfts should be the same. They will have metadata, metadata json, and current owners. Its not the collection we are fetching here but the nfts.

crates/core/src/db/queries/metadatas.rs
```diff
@@ -503,79 +509,124 @@ pub fn collection_nfts(conn: &Connection, options: CollectionNftOptions) -> Resu
...
+                    )
+                    .add(Expr::tbl(Listings::Table, Listings::Seller).equals(
+                        CurrentMetadataOwners::Table,
+                        CurrentMetadataOwners::OwnerAddress,
+                    )),
+            )
+            .and_where(Expr::col(Metadatas::BurnedAt).is_null())
+            .limit(limit)
+            .offset(offset)
+            .order_by((Listings::Table, Listings::Price), Order::Asc)
```
kespinola, line 565: The limit and offset will be the same for both queries consider assigning the base query to a variable then adding limit offset. 

Also order_by isn't always price. That will need some logic to add the order by based on the arguments to the function

kespinola (changes requested):
getting there. ty.

crates/core/src/db/queries/metadatas.rs
```diff
@@ -426,6 +445,230 @@ pub fn list(
...
+        order,
+        limit,
+        offset,
+    } = options;
+    let current_time = Utc::now().naive_utc();
+
+    let sort_unwrap = match sort_by.unwrap() {
+        Sort::Price => Listings::Price,
+        Sort::ListedAt => Listings::CreatedAt,
+    };
```
kespinola, line 490: This should be done with map and From trait.

```
let sort = sort_by.map_or(Listings::Price, Into::into)
```

This will map the Sort to Listing column or default to sorting by price.

It does require imp From<Sort> for Listings {}

crates/core/src/db/queries/metadatas.rs
```diff
@@ -426,6 +445,230 @@ pub fn list(
...
+
+    let order_unwrap = order.unwrap();
```
kespinola, line 492:
```suggestion
    let order_by = order.unwrap_or(Order::Desc)
```
Set desc as the default.

kespinola (changes requested):
Adjustments for handling sort and order defaults. Can you also look into resolving the clippy errors please?

kespinola (approved):

PR #747, opened 2022-09-15
Query collections ordered by floor price

alchemistgo87:
To be consumed by the trending collections UI - https://github.com/holaplex/dao-marketplace/issues/97


kespinola:
Closing because we now collection_trends table which tracks 1,7,30 vol, floor, and number of sales. 

We'll have 1 collectionTrends query.

PR #749, opened 2022-09-16
Delete GenoHabitats if exists before upsert

imabdulbasit:


crates/indexer/src/geyser/accounts/geno_habitat_data.rs
```diff
@@ -139,6 +156,23 @@ pub(crate) async fn process(
...
+            let geno_exists = select(exists(
+                geno_habitat_datas::table.filter(
+                    geno_habitat_datas::address
+                        .eq(key.to_string())
+                        .and(geno_habitat_datas::slot.lt(slot)),
+                ),
+            ))
+            .get_result::<bool>(db);
+
+            if let Ok(true) = geno_exists {
```
kespinola, line 168: Is the let necessary here since not using the left side down stream? Just use if.

kespinola (approved):

PR #758, opened 2022-09-19
Added STARTUP example  in env file

Maverick9081:
Added `startup` in env file when `STARTUP` variable is set to `null` in geyser config, so that indexer will connect to the correct rabbitmq queue


Maverick9081:
Stupid way of adding it , closing it

PR #761, opened 2022-09-20
Fix Genopets trigger

imabdulbasit:


imabdulbasit (commented):
already reran this migration on staging

kespinola (approved):

PR #775, opened 2022-09-22
Trending collections query

alchemistgo87:


crates/graphql/src/schema/query_root.rs
```diff
@@ -779,6 +782,41 @@ impl QueryRoot {
...
+            limit: limit.try_into()?,
+            offset: offset.try_into()?,
+        })?;
+
+        collections
+            .into_iter()
+            .map(TryInto::try_into)
+            .collect::<Result<_, _>>()
+            .map_err(Into::into)
+    }
```
kespinola, line 818: This needs to return a new object called CollectionTrend because the UI needs access to all the columns on the table.

CollectionTrend should have a field called collection which is either mcc or non-mcc collection.

crates/core/src/db/custom_types.rs
```diff
@@ -551,3 +551,35 @@ pub enum Sort {
...
+    #[strum(serialize = "_1dSalesCount")]
+    _1dSalesCount,
+
+    /// sort results by 7 days sales count
+    #[strum(serialize = "_7dSalesCount")]
+    _7dSalesCount,
+
+    /// sort results by 30 days sales count
+    #[strum(serialize = "_30dSalesCount")]
+    _30dSalesCount,
```
kespinola, line 584: Drop the _ from the enum but represent in its serializer.

crates/graphql/src/schema/context.rs
```diff
@@ -81,6 +81,7 @@ pub struct AppContext {
     pub nft_attributes_loader: Loader<PublicKey<Nft>, Vec<NftAttribute>>,
     pub nft_by_mint_loader: Loader<PublicKey<TokenMint>, Option<Nft>>,
     pub nft_collection_loader: Loader<PublicKey<Nft>, Option<Collection>>,
+    pub generic_collection_loader: Loader<String, Option<Collection>>,
```
kespinola, line 84:
```suggestion
    pub generic_collection_loader: Loader<CollectionIdentifier, Option<Collection>>,
```

@ray-kast attempted to to `pub CollectionIndentifier(String)` but got a good amount of trait errors from Loader. We are pressing on but a spot that could use a refactor.

PR #796, opened 2022-10-03
Fetch rewardcenter with auctionhouse

alchemistgo87:


crates/graphql/src/schema/context.rs
```diff
@@ -35,7 +35,7 @@ use scalars::{
     PublicKey,
 };
 
-use super::prelude::*;
+use super::{objects::reward_center::RewardCenter, prelude::*};
```
kespinola, line 38: Move the obeject import to the upper block on L6 `use objects::{...}`

kespinola (changes requested):
Import adjustment.

kespinola (approved):

Issue #813, opened 2022-10-11
Listing for a Collection Query


kespinola:
### Goal
Clients can query active listings for a collection in price ASC order.

```graphql
collection(id: "$collectionID") {
  listings(limit: 3, offeset: 0, auctionHouse: $auctionHouseAddress) {
    price
    nft {
      image
    }
  }
}
```
impl CollectionTrend {
 fn listings(&self, context) -> Vec<Listing> {
   ...perform query
 }
}

### Requirements
- Listings are sorted in price ASC order
- Listings scoped by auction house

### Tasks
- [x] Add index on listings (auctionHouse, price ASC)

PR #814, opened 2022-10-11
Add listing query for collection

alchemistgo87:
Addresses: #813

### DIFF
--- a/crates/graphql/src/schema/objects/nft.rs
+++ b/crates/graphql/src/schema/objects/nft.rs
@@ -927,7 +927,7 @@ pub struct CollectionTrend {
     pub thirty_day_marketcap_change: i32,
 }
 
-impl<'a> TryFrom<models::CollectionTrend> for CollectionTrend {
+impl TryFrom<models::CollectionTrend> for CollectionTrend {
     type Error = std::num::TryFromIntError;
 
     fn try_from(
--- a/crates/core/src/db/queries/listings.rs
+++ b/crates/core/src/db/queries/listings.rs
@@ -70,7 +70,7 @@ enum Listings {
     Expiry,
 }
 
-/// Input parameters for the [`collection_nfts`] query.
+/// Input parameters for the `collection_nfts` query.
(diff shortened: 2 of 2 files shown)


crates/core/src/db/queries/metadatas.rs
```diff
@@ -670,6 +683,148 @@ pub fn collection_nfts(conn: &Connection, options: CollectionNftOptions) -> Resu
...
+        .limit(limit)
+        .offset(offset)
+        .order_by((Listings::Table, Listings::Price), Order::Asc)
+        .take();
+
+    let query = query.to_string(PostgresQueryBuilder);
+
+    diesel::sql_query(query)
+        .load(conn)
+        .context("Failed to load nft(s)")
```
kespinola, line 825:
```suggestion
        .context("Failed to load collection listings")
```

This should be queries/listings and return Listing not NFTs. Still can make use sea query and the conditional to decide where to join in me_collections or metadata_collection_keys since need to filter out listings that don't belong to the collection.

Listings should then have a field called nft that uses the nft_loader to lookup the NFT for the listing.

crates/graphql/src/schema/objects/nft.rs
```diff
@@ -1147,6 +1150,28 @@ impl CollectionTrend {
...
+    pub async fn listings(
+        &self,
+        ctx: &AppContext,
+        limit: i32,
+        offset: i32,
+        auction_house: Option<String>,
+    ) -> FieldResult<Vec<Nft>> {
+        let conn = ctx.shared.db.get()?;
+
+        let nfts = queries::metadatas::collection_listed_nfts(&conn, CollectionListedNftOptions {
```
kespinola, line 1162:
```suggestion
        let nfts = queries::listings::list(&conn, CollectionListedNftOptions {
```

kespinola (changes requested):
To reduce complexity we should hold off on this until moonrank collections are in place that way we don't need to do the switching between mcc and me collections.

PR #824, opened 2022-10-13
Wallet offers query

alchemistgo87:


crates/core/src/db/queries/wallet.rs
```diff
@@ -8,14 +8,14 @@ use diesel::{
...
     db::{
-        models::{CollectedCollection, CreatedCollection, WalletActivity},
+        models::{CollectedCollection, CreatedCollection, Offer, WalletActivity},
         Connection,
     },
     error::prelude::*,
 };
 
 const ACTIVITES_QUERY: &str = r"
-SELECT listings.id as id, metadata, auction_house, price, auction_house, created_at, marketplace_program,
```
kespinola, line 18: Why the remove of auction_house?

crates/graphql/src/schema/objects/ah_offer.rs
```diff
@@ -36,6 +36,15 @@ impl Offer {
         &self.buyer
     }
 
+    async fn buyer_wallet(&self, ctx: &AppContext) -> Wallet {
+        let twitter_handle = ctx
+            .twitter_handle_loader
+            .load(self.buyer.clone())
+            .await
+            .unwrap_or_default();
```
kespinola, line 44: Is the unwrap_or_default needed? Wallet second parameter already accepts Option<String> which is the return of the loader.

crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -366,6 +366,24 @@ impl Wallet {
             .map_err(Into::into)
     }
 
+    pub fn offers(
+        &self,
+        ctx: &AppContext,
+        offer_type: Option<String>,
```
kespinola, line 372: Lets do a graphql enum type here. And then can have a from trait that converts to a string for the query.

kespinola (changes requested):
Some questions and request for enum type for offer_type. nj.

crates/core/src/db/queries/wallet.rs
alchemistgo87: Because it was written twice in the query.

crates/graphql/src/schema/objects/ah_offer.rs
alchemistgo87: Yes, otherwise it will send a result type and give the following compile time error:

```rust
mismatched types
expected enum `std::option::Option<_>`
   found enum `std::result::Result<std::option::Option<_>, batcher::Error>`
```

kespinola (approved):

mpwsh (approved):

Issue #842, opened 2022-10-21
Index v0.2.1 of Reward Center Program


kespinola:
### Goal
Index updated accounts and instructions for the reward center program.

```
RwDDvPp7ta9qqUwxbBfShsNreBaSsKvFcHzMxfBC3Ki
```

### Tasks
- [ ] Update program address
- [ ] Adjust account indexing based on new version of listing, offer, and remove of payment ticket.
- [ ] Index close_offer, close_listing, and execute_sale instructions
- [ ] Replace the canceled_at on offers and listings with canceled_at based on the time of indexing the close instructions
- [ ] Replace purchase_ticket with purchase_id to be based on the accounts and arguments of execute sale
- [ ] Write reward_payouts as part of execute_sale

### References
https://crates.io/crates/hpl-reward-center/0.2.1

PR #843, opened 2022-10-24
Reward Center 0.2.x Program Indexing

kespinola:
### Changes
- Update index logic to align with version 0.2.x of the program that closes accounts for offers and listings. Also the purchase_ticket account is removed. Instead relay on instruction indexing.

resolves https://github.com/holaplex/indexer/issues/842


crates/indexer/src/geyser/instructions/hpl_reward_center/execute_sale.rs
```diff
@@ -0,0 +1,165 @@
...
+    data: &[u8],
+    accounts: &[Pubkey],
+    slot: u64,
+) -> Result<()> {
+    let params = ExecuteSale::try_from_slice(data).context("failed to deserialize")?;
+
+    if accounts.len() < 21 {
+        debug!("invalid accounts for ExecuteSaleInstruction");
+        return Ok(());
+    }
```
kespinola, line 33: drop because need to pay out creators what is variable in number of accounts.

PR #857, opened 2022-10-28
Update reward program indexing to 0.2.3

kespinola:
### Changes
- Dropped passing in token size and price on close instructions for listing and offer since can read off of the account.

### DIFF
--- a/crates/indexer/src/geyser/instructions/hpl_reward_center/close_offer.rs
+++ b/crates/indexer/src/geyser/instructions/hpl_reward_center/close_offer.rs
@@ -29,39 +29,48 @@ pub(crate) async fn process(
 
     client
         .db()
-        .run(move |db| {
-            let (token_size, buyer_price) = rewards_offers::table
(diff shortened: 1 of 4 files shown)


crates/indexer/src/geyser/instructions/hpl_reward_center/close_offer.rs
```diff
@@ -25,43 +25,43 @@ pub(crate) async fn process(
...
         .run(move |db| {
-            insert_into(hpl_reward_center_close_offer_ins::table)
-                .values(&row)
-                .execute(db)
-        })
-        .await
-        .context("failed to insert reward center close offer instruction ")?;
+            let (token_size, buyer_price) = rewards_offers::table
+                .select((rewards_offers::token_size, rewards_offers::price))
+                .filter(rewards_offers::address.eq(offer_address.clone()))
```
kespinola, line 35: I moved all the inserts into the same run block because hitting move violations with offer_address.

PR #863, opened 2022-11-01
Fix wallet total reward - Return 0 if data does not exist

alchemistgo87:


crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -430,17 +430,22 @@ impl Wallet {
...
                 wallet_total_rewards::reward_center_address
                     .eq(&reward_center)
                     .and(wallet_total_rewards::wallet_address.eq(&self.address)),
             )
             .first::<BigDecimal>(&db_conn)
-            .context("Failed to load total rewards")?;
+            .context("Failed to load total rewards");
 
-        Ok(wallet_total_rewards.try_into().unwrap_or_default())
+        let wallet_total_reward = match wallet_total_reward {
```
kespinola, line 443: You can mark the result as optional() and the use unwrap default to write the zero.

kespinola (changes requested):

crates/graphql/src/schema/objects/wallet.rs
kespinola: And why big decimal isn't this i64? Should be no decimals.

crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -15,7 +15,10 @@ use scalars::{markers::TokenMint, PublicKey, U64};
 use tables::{associated_token_accounts, bids, graph_connections, wallet_total_rewards};
 
 use super::{ah_offer::Offer, prelude::*, reward_center::RewardCenter};
-use crate::schema::enums::{NftSort, OfferType, OrderDirection};
+use crate::schema::{
+    enums::{NftSort, OfferType, OrderDirection},
+    scalars::I64,
+};
```
kespinola, line 21: These can be accessed without the use crate::schema

use enums
use scalars

crates/graphql/src/schema/objects/wallet.rs
```diff
@@ -427,20 +430,20 @@ impl Wallet {
...
+        let wallet_total_reward = wallet_total_rewards::table
             .select(wallet_total_rewards::total_reward)
             .filter(
                 wallet_total_rewards::reward_center_address
                     .eq(&reward_center)
                     .and(wallet_total_rewards::wallet_address.eq(&self.address)),
             )
             .first::<BigDecimal>(&db_conn)
-            .context("Failed to load total rewards")?;
+            .unwrap_or_else(|_| BigDecimal::zero());
```
kespinola, line 444:
```rust
.optional()
.first::<BigDecimal>(&db_conn)
.unwrap_or(BigDecimal::zero())
```

This should be U64 why the switch to I64?

kespinola (changes requested):

crates/graphql/src/schema/objects/wallet.rs
alchemistgo87: This was after the last review. I supposed that it is required to be I64 - https://github.com/holaplex/indexer/pull/863#discussion_r1010360723

kespinola: Does this work

```rust
.first::<BigDecimal>(&db_conn)
.optional()
.unwrap_or(BigDecimal::zero())
```

alchemistgo87: This will throw error - `expected Option<BigDecimal>, found BigDecimal`

alchemistgo87: This is what actually worked:
```
.optional()
.transpose()
.unwrap_or_else(|| Ok(BigDecimal::zero()))
.unwrap_or_else(|_| BigDecimal::zero());
```

kespinola (approved):

PR #873, opened 2022-11-08
Sales (purchases) query

alchemistgo87:


kespinola (changes requested):
Is this for the hero section of nm? That should be query reward_payouts

PR #916, opened 2022-11-26
Index activities in separate table

imabdulbasit:


crates/core/migrations/2022-11-25-194809_create_collection_activities_table/up.sql
```diff
@@ -0,0 +1,11 @@
+create table if not exists collection_activities (
```
kespinola, line 1: Since we are deleting offers and listings the activity for nft and profiles isn't accurate. What do you think about calling this marketplace_activities.

For the profile query we query by either the buyer or seller
For nfts by the metadata
For collections by the collection id

crates/indexer/src/geyser/accounts/hpl_reward_center/listing.rs
```diff
@@ -157,6 +157,13 @@ pub(crate) async fn process(
                     return Ok(());
                 }
 
+                mutations::collection_activity::listing(
+                    db,
+                    listing_id,
+                    &listing.clone(),
+                    "LISTING_CREATED",
```
kespinola, line 164: Good for a rust enum

kespinola (changes requested):

kespinola (approved):
This is really great. Nj.

mpwsh (approved):

PR #935, opened 2022-12-08
Ignore meilisearch document with invalid id

imabdulbasit:


kespinola (approved):
This works but now we are skipping collections. Can we replace violations like a space with -a ? Can be a follow up.

mpwsh (approved):

PR #954, opened 2023-01-23
Get block time from Solana RPC for ME instructions

imabdulbasit:


mpwsh (approved):

kespinola (approved):

omarkilani:
I think this can be improved by adding a slot :: block time map (capped/LRU/MRU) to limit RPC load, if that's a concern.

PR #964, opened 2023-01-26
Fix activities indexing

imabdulbasit:


crates/geyser-consumer/src/handler/instructions/sell.rs
```diff
@@ -119,29 +119,30 @@ pub async fn upsert_into_listings_table<'a>(client: &Client, row: Listing<'stati
...
-            .get_result::<bool>(db)?;
+                )
+                .select(listings::slot)
+                .first(db)
+                .optional()?;
 
             let listing_id = mutations::listing::insert(db, &row)?;
 
-            if listing_exists {
+            if Some(row.slot) == indexed_listing_slot
```
kespinola, line 134: What is the case where we start a a backfill and the version of the listing is older than the 1 we have saved?

imabdulbasit: The slot_wv trigger would not update the listing but we record the activity

kespinola (approved):

PR #985, opened 2023-02-23
Use `magic_eden_id` in queries to Dolphin sources if available.

omarkilani:
This solves an issue with fetching dolphin data where indexer collection ids differ to those from dolphin (`smb` to `solana_monkey_business`, `abc` to  `abc_abracadabra`, etc)

### DIFF
--- a/crates/graphql/src/schema/objects/collection.rs
+++ b/crates/graphql/src/schema/objects/collection.rs
@@ -181,6 +181,15 @@ impl<'a> TryFrom<models::Collection<'a>> for Collection {
     }
 }
 
+impl Collection {
+    fn id_for_dolphin(&self) -> String {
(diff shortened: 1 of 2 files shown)


crates/graphql/src/schema/objects/collection.rs
```diff
@@ -181,6 +181,16 @@ impl<'a> TryFrom<models::Collection<'a>> for Collection {
     }
 }
 
+impl Collection {
+    fn id_for_dolphin(&self) -> String {
+        self.magic_eden_id
+            .as_deref()
+            .or(Some(&self.id))
```
kespinola, line 188: should look to return result instead of panicing with unwrap

kespinola (approved):

